PR #14, opened 2015-09-25
Fix sra tools

mvdbeek:
Various fixes for sra tools. See individual commits.


drosofff:
Can you add you email address in the help section https://github.com/mvdbeek/tools-artbio/blob/fix_sra_tools/tools/sra-tools/fastq_dump.xml#L139
and maybe the same in the help sections of other sra_tools ?

mvdbeek:
OK, added me and my mail address.

PR #19, opened 2015-09-28
Removed samtools from required dependencies for msp_sr_bowtie

drosofff:
and bump the version of the tool.


mvdbeek:
Thinking it through, I realize that the list of packages to be installed for galaxy has never been made official. Given that the tool may be run on a cluster without galaxy's core requirements, we should keep samtools as a dependecy for sRbowtie, because the wrapper makes direct calls to the samtools binary.

This is not true however for the remaining tools where we use pysam.
Could you instead uncomment the samtools dependency in the wrapper and tool_dependencies.xml?

PR #27, opened 2015-10-20
fetch_fasta_from_ncbi version 0.9.3

drosofff:
Polish the tool:
- bump to 0.9.3
- removing the experimental pubmed search which is not the purpose of the tool (yet I think it could be interesting to do text mining in galaxy)
- simplifying the test of fetched data integrity, in one single line test.
The version 0.9.3 has been tested on a huge download of nucleotide on mississippi. It is still running for proteins, but since 15 hours and it seems ok.


drosofff:
The version has passed a second test, although very adverse network conditions (both UPMC and NCBI sides).
I am merging myself the pull request

PR #38, opened 2015-11-11
Script for renaming history items and doc update to mod.py

mvdbeek:


drosofff:
@mvdbeek Can you add a doc to these scripts, explaining how to use and in which context ?

mvdbeek:
@drosofff for escape_for_rematch.py it's not worth the effort, I highly doubt we'll use it again.
I just renamed it from mod.py, because that doesn't mean anything. 
We could just as well remove the script, your call.

Does the demand for documentation it apply also for rename_hist_items.py ?
I was thinking the argparse was pretty OK. I can add an example tabular file.

PR #39, opened 2015-11-19
Make name and id configurable

mvdbeek:


mvdbeek:
@drosofff @fabiorjvieira with a commandline like this
`python get_tool_yml_from_gi.py -g https://mississippi.snv.jussieu.fr/ -a xxxxxx -l -skip_id -o tool_list.yml`
we get a tool_list yml that will allow placing newly installed tools using the latest version of https://github.com/galaxyproject/ansible-galaxy-tools  in the correct tool panel section !

PR #42, opened 2015-12-16
Metavisitor

drosofff:
Change the name of the suite_metavisitor_1_0 (previously suite_visitor2).
I will deprecate the suite_visitor2 on the main tool shed, after testing
Note also that a bump of msp_sr_readmap_and_size_histograms/readmap.xml to 1.0.5 had not been commited/merged


mvdbeek:
@drosofff 
Could you include package_cairo_1_12_14 by both devteam and iuc at the top of the repository dependencies? I'm suspecting a toolshed bug ... this would be a potential workaround.

drosofff:
should I accept or reject this PR ?

mvdbeek:
Accept, it was indeed a [toolshed bug](https://github.com/galaxyproject/galaxy/pull/1366)!

mvdbeek:
Though in the next iteration we can remove the explicit dependency on package_cairo_*

PR #45, opened 2016-01-11
Use travis for linting

mvdbeek:
and in the future use it for tests, too.


mvdbeek:
@drosofff can you merge if it's OK for you?

PR #47, opened 2016-01-22
create suite_metavisitor_1_2, which does not include ncbi_blast_plus …

drosofff:
…anymore

### DIFF
--- a/suites/suite_metavisitor_1_2/readme.rst
+++ b/suites/suite_metavisitor_1_2/readme.rst
@@ -1,7 +1,7 @@
 metavisitor
 
 version 1_2
-- The metavisitor suite now comes without the ncbi_blast_plus suite, which has to be installed separately from https://toolshed.g2.bx.psu.edu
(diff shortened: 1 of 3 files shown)


suites/suite_metavisitor_1_2/readme.rst
```diff
@@ -0,0 +1,12 @@
+metavisitor
+
+version 1_2
+- The metavisitor suite now comes without the ncbi_blast_plus suite, which has to be installed separately from https://toolshed.g2.bx.psu.edu
```
mvdbeek: ... seperately from https://toolshed.g2.bx.psu.edu/view/devteam/ncbi_blast_plus/7f3c448e119b ?

drosofff: Yes, let's give it a try: the idea is to not force the user to search for a particular changeset. This will oblige us to keep an eye on the workflows to see whether they are still compatible with next releases of ncbi+. But it make the suite lighter to install and clearer in terms of contributions.
thus I would say: separately from https://toolshed.g2.bx.psu.edu/view/devteam/ncbi_blast_plus/   ?
...

mvdbeek: Yes, perfect!

mvdbeek:
:+1:

PR #50, opened 2016-02-02
Add manual yrange option as requested by Antoine.

mvdbeek:


drosofff:
:+1: 
But I thought Antoine had also concerns with the range of the x axis ?

PR #51, opened 2016-02-03
Add add_input_name_as_column tool

mvdbeek:
@drosofff you may want to use `"$input.element_identifier" ` also for the metavisitor report issue.


drosofff:
And what happens if the input is _not_ a collection ?

mvdbeek:
> And what happens if the input is not a collection ?

It points to $inputfile.name, if I'm not mistaken.

PR #59, opened 2016-03-09
Goseq update

mvdbeek:
@drosofff @bgruening this is starting to be in decent shape.
I will split out the getgo tool into a seperate toolshed repo, since it needs all those organism packages.
Also I will need to work on the list of available genome / gene_id combinations for getgo.


drosofff:
gene_ids.loc and genome.loc are empty

mvdbeek:
they are symlinks.

tools/goseq_1_22_0/goseq.xml
```diff
@@ -1,62 +1,128 @@
-<tool id="goseq" name="goseq gene ontology analyser" version="0.1.0">
-    <description />
-    <requirements>
-        <requirement type="package" version="3.2.1">R</requirement>
-        <requirement type="package" version="1.22.0">goseq</requirement>
-    </requirements>
+<tool id="goseq" name="goseq" version="0.2.2">
+    <description>tests for overrepresented gene categories</description>
+    <macros>
+        <import>go_macros.xml</import>
```
drosofff: Well, maybe using the macros fits with the guide lines... But in this case this is the same number of code lines and does not really improve the readability.

mvdbeek: It does, it is also in the getgo and get_gene_length_and_gc_content xml.

tools/goseq_1_22_0/test-data/dge_list.tab
```diff
@@ -0,0 +1,1000 @@
+ENSG00000140459	False
+ENSG00000236824	False
+ENSG00000162526	False
```
bgruening: I would make the file small and use `contains`. I suspect this is changing from version to version and `contains` may also be more stable.

tools/goseq_1_22_0/get_length_and_gc_content.r
```diff
@@ -0,0 +1,59 @@
+# originally by Devon Ryan, https://www.biostars.org/p/84467/
+
+options( show.error.messages=F, error = function () { cat( geterrmessage(), file=stderr() ); q( "no", 1, F ) } )
+
+# we need that to not crash galaxy with an UTF8 error on German LC settings.
```
bgruening: Do you think this is better places in a config file?

bgruening:
This looks awesome! Do you still want to include it in IUC. Should I prepare a conda package?

tools/goseq_1_22_0/test-data/dge_list.tab
mvdbeek: this is an input file, but I guess for testing we could use less genes.

tools/goseq_1_22_0/get_length_and_gc_content.r
mvdbeek: well, I like that I can use these scripts on the commandline as well, and the debugging / development is much easier this way.

mvdbeek:
Yes, go ahead. All the functionality should be there.

tools/goseq_1_22_0/get_length_and_gc_content.r
bgruening: fair enough

bgruening:
Ok, done you have a PR. As soon as you move it to IUC it will be tested :)
Thanks @mvdbeek!

PR #64, opened 2016-03-09
Add conda dependency

bgruening:


mvdbeek:
Awesome, I'll see how it goes. I guess it should work for goseq and get_length_and_gc_content out of the box, for getgo there is this huge list of org.* packages. hg, dm and mm are already on bioconductor, I can try to get the remaining ones in.

mvdbeek:
Thanks @bgruening

PR #77, opened 2016-04-11
update of Galaxy-Workflow-Metavisitor__Workflow_for_Use_Case_3-2.ga

drosofff:
in the workflow toolshed repository


mvdbeek:
👍  do we have the corresponding update in https://github.com/artbio/ansible-artimed ?

drosofff:
Yes indeed ! this pull request is behind the one already merged in ansible-artimed. The diff returns no difference

drosofff:
And the workflow is also already updated in the toolshed by the way

PR #78, opened 2016-04-18
Exchange name with element_identifier for msp_sr_bowtie_parser

mvdbeek:
This is necessary so that we can recover the element name from dataset
collection elements.


drosofff:
well done 👍

PR #80, opened 2016-04-28
Skip failed k-mers with oases

mvdbeek:
Occasionally specific k-mers fail at random when assembling with oases/velvet.
If this is the case, we skip that k-mer and continue the assembly. Also bump
the tool version and and reformat wrapper xml to 4 space indentation, add
citation and fix help syntax errors.

### DIFF
--- a/tools/msp_oases/oases_optimiser.py
+++ b/tools/msp_oases/oases_optimiser.py
@@ -6,50 +6,48 @@ refactored using the adaptation of
 Konrad Paszkiewicz	University of Exeter, UK.
 
 """
-import pkg_resources;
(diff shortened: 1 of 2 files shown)


mvdbeek:
@drosofff @JuPeg I tested this with planemo, I guess this would make the tool more robust.

JuPeg:
I tested with datasets that were failing with previous version and now it works fine!

mvdbeek:
@drosofff should we give this a try? I think this will prevent a lot of frustration when running on larger datasets.

drosofff:
Yes, I will check also, and we need a list of what has to be updated in addition (tool list in ansible, toolshed suite, tool shed workflows, etc...). Start to be complicated to maintain isn't it ?

PR #86, opened 2016-05-23
Retrieve fasta from NCBI with bins

JuPeg:
Retrieving big databases from NCBI might fail due to a connection timeout. To solve this issue, the complete database is sliced in smaller requests. The bins are defined by delimitating the sequences length in the enter query, the ranges are defined by the user.

### DIFF
--- a/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
+++ b/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
@@ -263,15 +263,6 @@ def __main__():
     if len(args) > 0:
         parser.error('Wrong number of arguments')
 
-    try:
-        os.remove(options.outname)
-    except OSError:
-        pass
-    try:
-        os.remove(options.logfile)
-    except OSError:
-        pass
-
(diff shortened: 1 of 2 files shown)


JuPeg:
Apparently it just worked by luck. Testing again made me realise the error persists most of the times.

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -268,10 +270,19 @@ def __main__():
...
 
+    query=options.query_string
+    try:
+  	bins=map(int,options.bins.split(" "))
+	for i in range(len(bins)-1):
+        	options.query_string=query+" AND "+str(bins[i]+1)+":"+str(bins[i+1])+"[Sequence Length]"
+        	print options.query_string
+        	E = Eutils(options, logger)
+        	E.retrieve()
+    except:
```
mvdbeek: Do not use bare Exception when possible. Usually you can test for the specific Exception like so
(KeyError would be the exception for a non-existing key in a dictionary)
```
except KeyError:
```
You can see the correct key in the traceback.

PR #87, opened 2016-07-09
Add element identifer to size_histogram and bump version

mvdbeek:
ping @drosofff


drosofff:
The version is pushed in the toolshed anyway. Is it best practice ?

PR #88, opened 2016-07-15
Catch errors at retrieve from NCBI

JuPeg:
I catch exceptions to every error I have found while testing. 
Sometimes no error is thrown but instead of sequences we retrieve error messages. In this case, we also try to retrieve the same batch again. 
I only got the connection reset by peer once, so I just break it, as I couldn't test how to fix it.

### DIFF
--- a/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
+++ b/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
@@ -169,7 +169,7 @@ class Eutils:
             try:
                 response = urllib2.urlopen(req)
                 fasta = response.read()
-                if ("Resource temporarily unavailable" in fasta) or (not fasta.startswith(">") ):
+                if ("Resource temporarily unavailable" in fasta) or ("Error" in fasta) or (not fasta.startswith(">") ):
                     serverTransaction = False
                 else:
                     serverTransaction = True
--- a/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
+++ b/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
@@ -33,7 +33,6 @@ import urllib2
 import httplib
 import re
 from socket import error as SocketError
-from datetime import datetime
 
 class Eutils:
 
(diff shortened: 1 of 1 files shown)


tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -166,16 +169,30 @@ def efetch(self, db, query_key, webenv):
...
             except urllib2.HTTPError as e:
+                print "urllib2.HTTPError at ", datetime.now()
                 serverTransaction = False
                 self.logger.info("urlopen error:%s, %s" % (e.code, e.read() ) )
             except httplib.IncompleteRead as e:
+                print "httplib.IncompleteRead at ", datetime.now()
+                serverTransaction = False
+                self.logger.info("IncompleteRead error")
+            except urllib2.URLError as e:
+                print "urllib2.URLError at ", datetime.now()
```
mvdbeek: I'm unsure how the logger is configured, but it includes the date by default, so unless the logger is silent, the print statements don't add any information.

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -166,16 +169,30 @@ def efetch(self, db, query_key, webenv):
             try:
                 response = urllib2.urlopen(req)
                 fasta = response.read()
```
mvdbeek: I think a better solution is to use
```
if response.getcode() == 200:
    fasta = response.read()
    serverTransaction = True
else:
    continue
```

Normally that should cover all the exceptions you're trying to catch later on,
but I haven't tested this.

mvdbeek:
Looks like a nice addition!

mvdbeek:
It seems that you can provoke connection reset by peer with tcpkill. See https://stackoverflow.com/questions/3139204/triggering-connection-reset-by-peer

drosofff:
@JuPeg if do the proposed changes and commit into this branch, the test should work now

JuPeg:
I didn't have time to test much, but using response.getcode() seems to work fine.

PR #90, opened 2016-09-18
Update required bowtie package to reach parity with conda

mvdbeek:
planemo test still gives same test-result, and changes were minimal between 0.12.7 and 1.x.


drosofff:
👍 thanks

drosofff:
@mvdbeek 
Correct me if I am wrong: the updated package does not install properly.
Tested on plastisipi (16_07) and mississippi (16_04) :
```

Name	Version	Type	Status	Error
bowtie	1.1.2	package	Error	/bin/sh: g++: command not found
make: *** [bowtie-build-s] Error 127
```

mvdbeek:
`g++` should really be installed on the servers.
`apt-get install build-essential ` or `apt-get install g++` should be enough.

PR #91, opened 2016-10-07
Pin r version

mvdbeek:
The release of r-3.3.1 in conda has caused some problems when resolving conda dependencies.
I would say fixing the r-version is a good idea.

In addition this PR adds iuc-style testing of tools (copied from https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml).
If a tool is supposed to be tested, remove it from the .tt_blacklist file (and modify the tool, so that it gets picked up as having changes).


mvdbeek:
The good news is that testing works, and we find errors:
```
https://travis-ci.org/ARTbio/tools-artbio/jobs/165766611#L3199
```
The bad news is that I don't really know why, local testing on OS X works.
But that's the whole point of testing.

mvdbeek:
Okay! https://travis-ci.org/ARTbio/tools-artbio/jobs/165766611#L3181, it didn't find r-latticextra.

mvdbeek:
@drosofff Alright, we have a passing test https://travis-ci.org/ARTbio/tools-artbio/jobs/166019882#L2953

drosofff:
OK thanks. What about the development workflow ? : 
Should we now modify the tt_blacklist (since the test has passed) and include it to this PR before merging ? Or a tool PR will be accompanied by another PR after merging, with just the tt_blacklist change ?

mvdbeek:
If the tool is new, i.e it's not in the blacklist, write a test for the tool as usual and make sure it passes. If the tool is on the blacklist, remove it from the blacklist and touch the tool xml (just change the readme for example), so that planemo realises that the tool has been changed.

.travis.yml
```diff
@@ -1,12 +1,47 @@
...
-  - planemo --version
-  - planemo shed_lint --report_level warn --tools --fail_level error --recursive tools/
-  - planemo shed_lint --report_level warn --tools --fail_level error --recursive packages/
-  - planemo shed_lint --report_level warn --tools --fail_level error --recursive suites/
+  - cd "$TRAVIS_BUILD_DIR"
+  - while read -r DIR; do planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR" || exit 1; done < changed_repositories_chunk.list
+  - while read -r DIR; do planemo conda_install "$DIR"; done < changed_repositories_chunk.list
+  - while read -r DIR; do planemo test --conda_dependency_resolution --galaxy_branch "$GALAXY_RELEASE" --galaxy_source "$GALAXY_REPO" "$DIR" || exit 1; done < changed_repositories_chunk.list
+
+#after_success:
```
mvdbeek: @drosofff we could also automatically upload the tools to the toolsheds by uncommenting the following lines and setting SHED_PASSWORD and SHED_EMAIL in the travis settings. The IUC is doing it like this. We could maybe try doing this for the testtoolshed and see how it goes?

drosofff:
> make sure it passes
... apparently you did not make enough for this tool :--)

No, but seriously, you did not answer. here, now that the tool looks ok, the next step is too blacklist it. So this will be another PR, right ?

mvdbeek:
> No, but seriously, you did not answer. here, now that the tool looks ok, the next step is too blacklist it. So this will be another PR, right ?

The blacklist is for tools that should not be tested (e.g. because no conda package is available).
The idea is to take the existing tools off the blacklist one by one.

.travis.yml
drosofff: In principle yes. I'll keep it in the todo list

drosofff:
OK. I did not look carefully to the test code:

Tools that are not in the blacklist are tested **IF** _their_ code of is affected by the PR, is that correct ?

I guess (hope) that not all conda valid tool are not retested each time a single tool is updated. You see what I mean ?

mvdbeek:
> Tools that are not in the blacklist are tested IF their code of is affected by the PR, is that correct ?

yes

> I guess (hope) that not all conda valid tool are not retested each time a single tool is updated. You see what I mean ?

only tools that have been changed in the commit range will be tested, so if you look at https://github.com/ARTbio/tools-artbio/pull/92 you see only cherry_pick_fasta and add_input_name_as_column get tested, and not msp_sr_readmap_and_histogram.

PR #92, opened 2016-10-08
Test add_input_name_as_column and cherry_pick_fasta tools

mvdbeek:
As an example of how we can reduce the blacklist and increase the amount of tools under testing.


drosofff:
OK, I see. 👍

PR #93, opened 2016-10-09
Update majority of tools for conda & automatic testing

mvdbeek:
This moves all tools that have conda packages available (or that don't have any dependency) under testing.
While I am at it, I also corrected a subset of flake8 issues, most notably we're avoiding `from smRtools import *`.

I took care that most tools keep conda and TS dependencies when possible.
Conda-only tools are now mismatch_frequencies , get_sra_info and the goseq tools, which I think I am the only user. These tools had dependency problems anyway.


tools/msp_blastparser_and_hits/BlastParser_and_hits.xml
```diff
@@ -1,12 +1,12 @@
...
 <description>for virus discovery</description>
 <requirements></requirements>
-<command interpreter="python">
-BlastParser_and_hits.py
-	--sequences $sequences
-	--blast $blast
-	--tabularOutput $tabularOutput
-	--fastaOutput $fastaOutput
+<command><![CDATA[
+    python '$_tool_directory__'/BlastParser_and_hits.py
```
drosofff: python '$__tool_directory__'/

mvdbeek: i know, and there are a few other things in this wrapper that are not correct.
The [WIP] label means work in progress, I will change this when its done and a review is helpful.

mvdbeek:
Alright @drosofff, I think this is ready now.

drosofff:
testing does not work 27 days latter, with sqlite3.OperationalError

mvdbeek:
The sqlite errors are random hiccups with galaxy amd went away when testing again. That's why we use postgresql in production.
The other failure is a change in the fasta header returned by the NCBI, so that's out of our control. Testa are passing now :).

drosofff:
AH ok, very good, thanks 👍

PR #95, opened 2016-11-09
update changeset revisions tags of the suite_metavisitor_1_2

drosofff:
The build fails but I guess it's just because actually no tools were updated, just the tool suite definition


mvdbeek:
Actually the pr build worked, while the push build failed.This is either a planemo issue or a problem with
$TRAVIS_COMMIT_RANGE. But this is a safe change anyway. If you want to explore this I would be curious if this also happens if you try this PR from a fork.

drosofff:
I will be ready to merge this afternoon, toolshed is updated

PR #98, opened 2016-12-06
Fix travis yml

drosofff:
the current .travis.yml differs only from that one of iuc-tools by 
- && flake8 --exclude=.git .
- --urls


drosofff:
follow up on the two diffs

PR #99, opened 2016-12-07
Lumpy

drosofff:
I first decent release of my lumpy wrapper searching for structural variations


tools/lumpy/lumpy.xml
```diff
@@ -0,0 +1,261 @@
...
+        <requirement type="package" version="1.3.1">samtools</requirement>
+        <requirement type="package" version="1.11.2">numpy</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" level="fatal" description="Tool exception" />
+    </stdio>
+    <version_command>lumpy --version</version_command>
+    <command><![CDATA[
+        #if $analysis_type.analysis_type_list == "one_sample":
+            ln -f -s $analysis_type.input_file input.bam &&
```
mvdbeek: Paths should generally be quoted with single quotes.

tools/lumpy/lumpy.xml
```diff
@@ -0,0 +1,261 @@
+<tool id="lumpy" name="lumpy-sv" version="0.2">
+    <description>find structural variants</description>
+    <requirements>
+        <requirement type="package" version="0.2.12">lumpy-sv</requirement>
+        <requirement type="package" version="1.3.1">samtools</requirement>
+        <requirement type="package" version="1.11.2">numpy</requirement>
```
mvdbeek: Since the wrapperscript doesn't look like it's python3 compatible I would fix the numpy version to py27.
I think `1.11.2=py27` in the version tag should do it.

tools/lumpy/lumpy.xml
```diff
@@ -0,0 +1,261 @@
...
+                    <param name="min_mapping_threshold" value="20"  type="integer" label="min_mapping_threshold" help="e.g. 20" />
+                </section>
+            </when>
+            
+        </conditional>
+
+    </inputs>
+
+    <outputs>
+        <data format="tabular" name="histogram" label="Lumpy on ${analysis_type.input_file.element_identifier}: Fragment size distribution">
```
mvdbeek: Have you tested this with collections? I am not sure this would give you a reasonable name?
Perhaps ${on_string} would be better?

tools/lumpy/lumpy.xml
```diff
@@ -0,0 +1,261 @@
...
+	-e	Show evidence for each call
+	-w	File read windows size (default 1000000)
+	-mw	minimum weight for a call
+	-msw	minimum per-sample weight for a call
+	-tt	trim threshold
+	-x	exclude file bed file
+	-t	temp file prefix, must be to a writeable directory
+	-P	output probability curve for each variant
+	-b	output BEDPE instead of VCF
+	-sr	bam_file:&lt;file name&gt;,
```
mvdbeek: you could wrap the help text in a <![CDATA[]]> tag, so use `>` or `<` without escaping.

tools/lumpy/pairend_distro.py
```diff
@@ -0,0 +1,140 @@
...
+else:
+    # Remove outliers
+    L = np.array(L)
+    L.sort()
+    med, umad = unscaled_upper_mad(L)
+    upper_cutoff = med + options.mads * umad
+    L = L[L < upper_cutoff]
+    new_len = len(L)
+    removed = c - new_len
+    sys.stderr.write("Removed %d outliers with isize >= %d\n" %
```
mvdbeek: I would add `<command detect_errors="exit_code">`, so that this doesn't result in red datasets

tools/lumpy/lumpy.xml
drosofff: like this `'ln -f -s $analysis_type.input_file input.bam'` ?

drosofff: oops.. mean `ln -f -s '$analysis_type.input_file input.bam'`

tools/lumpy/pairend_distro.py
drosofff: In the xml file ?

tools/lumpy/lumpy.xml
mvdbeek: sorry, was the github UI swalling the tag. it's `<![CDATA[]]>`

tools/lumpy/pairend_distro.py
mvdbeek: in the command tag in line 12 of the wrapper xml, yes.

drosofff:
Is BWA-mem accepting list of dataset pairs, or paired list ? I feel there is maybe a problem with the tool that does not reduce properly to a single BAM (or does not map correctly the inputs).
I am trying to avoid fixing something in lumpy wrapper that is not an issue, but the issue of BWA-mem

mvdbeek:
> Is BWA-mem accepting list of dataset pairs, or paired list ?

I'm using 0.7.12.1, there it works fine, I'm using this in a workflow.
Make sure you select `Paired Collection` in the `Single or Paired-end reads` select field.
If the input is a list of paired collection (`list:paired`), the output will be a flat dataset collection (`list`).

drosofff:
@mvdbeek yes, I missed this option in the bwa tool form. It works like a charm now, thanks

PR #102, opened 2016-12-10
Fix hardcoded readgroups

mvdbeek:
See the commits for details.

### DIFF
--- a/tools/lumpy/lumpy.xml
+++ b/tools/lumpy/lumpy.xml
@@ -8,7 +8,6 @@
     <stdio>
         <exit_code range="1:" level="fatal" description="Tool exception" />
     </stdio>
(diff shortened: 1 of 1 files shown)


mvdbeek:
If you want to verify this, you can set the readgroup in the input test data to the filename like this:
```
samtools merge -r - test-data/sr.input.bam > test.bam
```
now if you do `samtools view -r readgroup test-data/sr.input.bam` you get the bam printed to screen,
but if you do the same with the new test.bam file you get nothing, because the readgroup readgroup is not in the file.
If you do `samtools view -r sr.input test.bam` you will get the file printed to the screen.

Let me just rebase this quckly.

drosofff:
Yeah, that's much better like that ! I had also completely forgot my hard coded 101 readlength...

PR #106, opened 2016-12-13
Add tests, pipe where possible, remove tail and add -e and -P options

mvdbeek:
I understand the `tail`  command is in there to work around a broken pip for samtools,
but I'm still getting this error. It's not a "real" problem, it's just that the pairend_distro.py script closes the pipe (=by quitting) once the desired sample has been taken.


tools/lumpy/lumpy.xml
```diff
@@ -9,64 +9,53 @@
...
+            ln -f -s '$analysis_type.input_file' '$one_sample_bam' &&
         #else:
-            ln -f -s '$analysis_type.input_file' input.A.bam &&
-            ln -f -s '$analysis_type.input_fileB' input.B.bam &&
+            #set sample_a_bam = re.sub('[^\w\-]', '_', str($analysis_type.input_file.element_identifier))
+            #set sample_b_bam = re.sub('[^\w\-]', '_', str($analysis_type.input_fileB.element_identifier))
+            #if $sample_a_bam == $sample_b_bam:
+                #set sample_a_bam = "%s_a" % str($sample_a_bam)
+                #set sample_b_bam = "%s_b" % str($sample_b_bam)
+            #end if
```
drosofff: Correct me if I am wrong: if $sample_a_bam == $sample_b_bam in this block, the user made a mistake (input_file.element_identifier and input_fileB.element_identifier received the same value). I am not saying that we should not cope with that, but if on the other hand, the purpose of this is a twist to get the test for 2 samples running with a single test file, this is weird ! Please explain. On a side note, for my knowledge of cheetah/python: the str() fonction is required ? I thought it was the default type in this case.

tools/lumpy/lumpy.xml
```diff
@@ -9,64 +9,53 @@
...
         #if $analysis_type.analysis_type_list == "one_sample":
 
             #if $seq_method.seq_method_list == "paired-end":
-                samtools view -b -F 1294 input.bam > "input.discordants.unsorted.bam" &&
-                samtools view -h input.bam | python $__tool_directory__/extractSplitReads_BwaMem.py -i stdin | samtools view -Sb - > "input.splitters.unsorted.bam" &&
-                samtools sort input.discordants.unsorted.bam > input.discordants.bam &&
-                samtools sort input.splitters.unsorted.bam > input.splitters.bam &&
-                samtools view input.bam
-                    |tail -n +1
+                samtools view -F 1294 '$one_sample_bam' | samtools sort -O bam -o input.discordants.bam - &&
```
drosofff: Is the last - of the CL required ? I never understand exactly when it is or not.
for instance cat some_file | cat returns the same output as cat some_file | cat -

tools/lumpy/lumpy.xml
```diff
@@ -160,21 +135,21 @@
     </inputs>
 
     <outputs>
-        <data format="tabular" name="histogram" label="Lumpy on ${on_string}: Fragment size distribution">
+        <data format="tabular" name="histogram" label="Lumpy on ${on_string}: Fragment size distribution" from_work_dir="input.lib.histo">
```
drosofff: Again "for my culture": what happens to files that are in the from_work_dir but are not collected through this tag ? No need to clean it with an rm ? (I see you do not use this command anymore)

tools/lumpy/lumpy.xml
mvdbeek: > Can you explain the purpose on this code expansion please ?

The above part makes sure we get correct names in the evidence column. That should help when you combine the VCFs from multiple comparisons.
This part here just makes sure that the sample identifiers are different (which is not essential, I can remove this if you prefer).

> I am not saying that we should not cope with that, but if on the other hand, the purpose of this is a twist to get the test for 2 samples running with a single test file, this is weird !

Just think of our bowtie wrapper, which always produces the output "Bowtie Output". I would consider this defensive programming. FWIW, you can't produce collections that have the same element identifier. Yet, for someone that just quickly throws in 2 datasets to check that the wrapper is working I think this is essential. If this was just to get the test data going I would be better of symlinking the same file as file_a.bam and file_b.bam. Your call.

tools/lumpy/lumpy.xml
mvdbeek: I just followed the samtools helptext, which says `Usage: samtools sort [options...] [in.bam]`.
But it also works without the `-`.

mvdbeek:
> Just that I don't like this in the stderr : for people who do not know the tool from inside (a majority I guess), it is stressing

Absolutely, I was also worried about this. I think we could possibly fix this by calling `samtools view` directly from within pairend_distro.py, or to just drop samtools stderr with`2> /dev/null`.

drosofff:
add -e and -P options ?

mvdbeek:
If you select thos you get a more detailed VCF output:
```
-e       Show evidence for each call
...

-P       output probability curve for each variant
```
If you want to see the difference you can compare the output_extended.vcf (where both -e and -P are selected) with output.vcf in the test-data folder.

mvdbeek:
Ahh, now I get your question -- I forgot to add the changes in the wrapper.

drosofff:
I tested the PR with planemo serve and my datasets (paired-end) and it fails. The same datasets pass the test with the current master branch, of course.
The current tests encoded in the tool are not accurate enough and the planemo test by Travis is not sufficient to ensure the code is working

drosofff:
FYI, broken code is introduced in the commit b929003809d6602f334ce5dab1f25104727c8355. Parent commit works for me

mvdbeek:
`-b` (output as bam (instead of sam)) shouldn't be necessary ... it outputs whatever format is the input, but samtools sort can deal with both. From `samtools view`:
```
  6. Option `-u' is preferred over `-b' when the output is piped to
     another samtools command.
```
So this should be changed to -u. I guess something else must be wrong in addition. I will add some paired-end test data.

drosofff:
>  I will add some paired-end test data.

In a separate PR please. I will not take it in this PR.

drosofff:
Closing this PR. Follow up in PR #107

PR #108, opened 2016-12-18
reintegrate --urls for planemo --shed_lint in travis yml and add comment about flake8

drosofff:


.travis.yml
```diff
@@ -39,8 +39,8 @@ install:
   - cat changed_repositories_chunk.list
 
 script:
-  - cd "$TRAVIS_BUILD_DIR"
-  - while read -r DIR; do planemo shed_lint --tools --ensure_metadata --report_level warn --fail_level error --recursive "$DIR" || exit 1; done < changed_repositories_chunk.list
+  - cd "$TRAVIS_BUILD_DIR" # && flake8 --exclude=.git . # no black list for python linting ?
```
mvdbeek: `--exclude` can be expanded with comma-separated values.
[see](http://flake8.pycqa.org/en/latest/user/configuration.html)
If you want to exclude for example everything in the unstable folder you could change this to
flake8 --exclude=.git,unstable/

PR #113, opened 2017-03-22
remove data_manager from .travis.yml

chamaelj:


drosofff:
👍 thanks we do not have datamanager tools indeed

.travis.yml
```diff
@@ -32,7 +32,6 @@ install:
   - echo $TRAVIS_COMMIT_RANGE
   - |
       planemo ci_find_repos --exclude_from .tt_blacklist \
-                            --exclude packages --exclude data_managers \
```
drosofff: You have to keep `--exclude packages` because we have packages in tool-artbio !

drosofff:
`--exclude packages` is duplicated now ;-) and apparently planemo ci_find_repos does not like it...

drosofff:
👍

PR #117, opened 2017-04-15
Fixoases

drosofff:
The tool now uses exclusively relies on bioconda velvet and oases dependencies. tool_dependencies.xml file is removed.
As with the oases conda packages the output appears to not be deterministic anymore, the test now relies on the presence of expected fasta headers in oases assembled transcripts.
The README is updated accordingly, and the tool version is bumped to 1.2.0


drosofff:
fixing .travis.yml with PR #119 seems to solve the test of msp_oases. I will remove again the output.fa test file since it should not be required anymore

PR #124, opened 2017-05-09
Smallrnamap

chamaelj:


drosofff:
@chamaelj it's not crystal clear to me why you need the .bai file to get your script working. Can you elaborate on this ?
Otherwise, take a look to https://www.biostars.org/p/15847/ . Indeed - if you really need this .bai file, you can generate it internally with samtools index file.bam. But yet, I am not sure your really need it. Even the lumpy wrapper did not need the .bai file, just a *sorted* bam file.

tools/small_rna_map/small_rna_map.py
```diff
@@ -0,0 +1,82 @@
...
+			polarity = "F" if (not read.is_reverse) else "R"
+                        read_rname = read.reference_name
+                        read_position = read.pos + 1
+                        the_key = (read_rname, read_position,polarity)
+                        try:
+                                ref_coord[the_key] = ref_coord[the_key] + 1
+                                
+                        except:
+                                ref_coord[the_key] = 1
+        return collections.OrderedDict(sorted(ref_coord.items())) #dictionary {(read_rname,read_position,polarity):nmbr_read}
```
drosofff: I don't see why you need an OrderedDict as you do not manipulate it afterward. It's not really important to sort the dictionary at this stage.

tools/small_rna_map/small_rna_map.py
```diff
@@ -0,0 +1,82 @@
...
+	return the_name[0]
+
+
+def main(infile, output):
+        with pysam.AlignmentFile(infile,"rb") as infile:
+                with open(output,"w") as outfile:
+                        outfile.write("Dataset" + "\t" + "Chromosome"+ "\t" + "Chrom_length" + "\t" + "Coordinate" + "\t" + "Nbr_reads" + "\t" + "Polarity" + "\t" + "Max" + "\t" + "Mean" + "\t" + "Median" + "\n")
+                        ref_coord = Ref_coord_pol(infile)
+                        ref_length = Ref_length(infile)
+                        for read in ref_coord:
```
drosofff: here did you try just:
```
for read in sorted(ref_coord):
```
?

another point, did you test how yous tuple keys are indeed sorted ? --> I would rather code the tuple like
```
(read_rname, polarity, read_position)
```
for clarity of the output, but it's a detail, question of taste...

tools/small_rna_map/small_rna_map.py
```diff
@@ -0,0 +1,82 @@
...
+				outfile.write(str(filename(infile)) + "\t" + str(read[0]) + "\t" + str(ref_length[str(read[0])]) + "\t" + str(read[1]) + "\t" + str(ref_coord[read])+ "\t" + str(read[2]) + "\t" + str(calcul_stat(ref_coord,read[0])[0]) + "\t" + str(calcul_stat(ref_coord,read[0])[1]) + "\t" + str(calcul_stat(ref_coord,read[0])[2])+ "\n")
```
drosofff: There is most likely a more efficient way to do that, based on outfile.write("\t".join(list_of_your_items)),
See also whether the write method has something to add "\n" at the end of the argument. I would be surprised that there is nothing but "+ "\n". In general the + concatenation is poorly efficient because it consume variables that are non-mutable.

drosofff: And "Oh la la".... are you going to re-run the calcul_stat function for each key of the ref_coord dictionary ??? Maybe I missed something but it sounds like over-redundant... or you missed something on the mean size of reads at a given coordinate

tools/small_rna_map/small_rna_map.xml
```diff
@@ -0,0 +1,39 @@
+<tool id="small_rna_map" name="small rna map" version="0.0.1">
+  <description></description>
+  <requirements>
+        <requirement type="package" version="3.1.2">R</requirement>
+        <requirement type="package" version="1.3.0=r3.2.2_1">r-optparse</requirement>
```
drosofff: You also need to specify numpy in your requirements as in https://github.com/ARTbio/tools-artbio/blob/master/tools/lumpy/lumpy.xml#L6 and probably ggplot for the future (look at bioconda if ggplot package is available)

tools/small_rna_map/test-data/output.tab
```diff
@@ -0,0 +1,465 @@
+Dataset	Chromosome	Chrom_length	Coordinate	Nbr_reads	Polarity	Max	Mean	Median
+dataset_1	FBtr0070001	72	17	1	F	17	17.0	17.0
```
drosofff: Apparently you have lost the sorting on the way, or it's an unexpected sorting

tools/small_rna_map/small_rna_map.py
drosofff: My bad, you're right actually !
Yet the structure of the dictionary should be {(read_rname,read_position,polarity):[readsize1, readsize2, ...]}
had the end, the length of the list gives you the number of reads, and you can easily compute statistics on the list (median, mean of the read lengths at a given position)

tools/small_rna_map/small_rna_map.py
```diff
@@ -35,28 +35,28 @@ def Ref_coord_pol(bamfile):
                         read_rname = read.reference_name
                         read_position = read.pos + 1
                         the_key = (read_rname, read_position,polarity)
+                        cigar = read.cigar[0][1]
```
drosofff: I think this method .cigar is tagged as deprecated (http://pysam.readthedocs.io/en/latest/api.html)
In addition are you sure that is always return the read **length** ?
Why not using the pysam method query_length in the class pysam.AlignedSegment ? -- pysam.AlignedSegment.query_length

drosofff: On another note: be simple, if cigar is read length call it read_length, not cigar.

drosofff:
👍

tools/small_rna_map/small_rna_map.r
```diff
@@ -0,0 +1,40 @@
...
+args = parse_args(parser)
+
+theme_set(theme_bw())
+Table = read.delim(args$output_tab, header=T, row.names=NULL)
+
+for(i in 1:nrow(Table)){
+  if (Table$Polarity[i] == "R"){
+	Table$Nbr_reads[i]= -1 * Table$Nbr_reads[i]
+	}
+}
```
drosofff: R loops are very slow. Use R function whose "internal" looping is implemented in C:
lines 16 to 20 can be replaced by
Table <- within(Table, Nbr_reads[Polarity=="R"] <- (Nbr_reads[Polarity=="R"]*-1))

tools/small_rna_map/test.r
```diff
@@ -0,0 +1,15 @@
+# Table is the data frame
+p <- ggplot(Table, aes(x=`Nbr_reads`, y=Coordinate, label=Nbr_reads)) +
```
drosofff: to simplify the test, only use several chromosomes like this:

p <- ggplot(Table[1:27,], aes(x=`Nbr_reads`, y=Coordinate, label=Nbr_reads)) +

You will then easily see that values for both x-axis and y-axis are not good. Also, remote the black dots as well as the nbr_reads labels in these dots. the information is redundant for the moment and complicates the development

drosofff:
@chamaelj I will close this pull request:

I have created a smallrnamap branch on the ARTbio/tools-artbio repo.
I'll show you how you can now push your developments in this remote branch. The benefit is that we can more easily work together on the same code. The price is that you have to regularly pull this branch in you local development repo to keep track of the changes you are not responsible for.

PR #133, opened 2017-06-01
Cleaning retrieve fasta from ncbi

ccedmendoza:
Bumped script to 1.1 and added a way to handle Error in fasta so the script doesn't loop infinitely

### DIFF
--- a/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
+++ b/tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
@@ -195,6 +195,14 @@ class Eutils:
                 if ( (response_code != 200) or ("Resource temporarily unavailable" in fasta)
                     or ("Error" in fasta) or (not fasta.startswith(">") ) ):
                     serverTransaction = False
+                    if ( response_code != 200 ):
+                        self.logger.info("urlopen error: Response code is not 200")
+                    elif ( "Resource temporarily unavailable" in fasta ):
+                        self.logger.info("Ressource temporarily unavailable")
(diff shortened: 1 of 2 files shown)


tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -271,7 +283,10 @@ def get_sequences(self):
                     while not mfasta:
                         self.logger.info("retrieving batch %d" % ((start / batch_size) + 1))
                         mfasta = self.efetch(self.dbname, self.query_key, self.webenv)
-                    out.write(mfasta + '\n')
+                    if ("counter" not in mfasta):
+                        out.write(mfasta + '\n')
+                    else:
+                        self.logger.info("Server Trials numbered more than 500. Skipping this batch")
```
drosofff: It would be interesting to keep this batch in raw format, to have an idea about what has been skipped (for debugging). If some NCBI entries are invalid, better to blacklist them

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -218,7 +227,7 @@ def efetch(self, db, query_key, webenv):
             if (counter > 500):
                 serverTransaction = True
         if (counter > 500):
-            return "counter"
+            raise QueryException({"message":"500 Server Transaction Trials attempted for this batch. Guiving up."})
```
drosofff: Giving up. 
But I would favor just "Aborting"

PR #150, opened 2017-07-11
Move msp_sr_size_histograms (drosofff) to small_read_size_histograms (artbio)

drosofff:
- create new repo for the tool now owned by artbio
- upgrade dependencies versions
- separate R script from the xml wrapper (no more config file)
- use r-optparse for passing parameters to the R script


tools/small_read_size_histograms/size_histogram.xml
```diff
@@ -1,15 +1,14 @@
...
     <requirements>
-        <requirement type="package" version="1.1.2">bowtie</requirement>
-        <requirement type="package" version="0.8.3">pysam</requirement>
-        <requirement type="package" version="3.1.2">R</requirement>
-        <requirement type="package" version="2.14">biocbasics</requirement>
-        <requirement type="package" version="1.9">numpy</requirement>
-        <requirement type="package" version="2.0.0=r3.2.2_0a">r-gridextra</requirement>
-        <requirement type="package" version="0.6_26=r3.2.2_2a">r-latticeextra</requirement>
+        <requirement type="package" version="1.2.0=py27_0">bowtie</requirement>
+        <requirement type="package" version="0.11.2.1=py27_0">pysam</requirement>
```
ccedmendoza: You said you wanted to slowly change the python version of the scripts in Galaxy to python3.
Wouldn't it be better to do it here instead of leaving it in python2 then changing it to python3 an other time?
(I ask this because this question applies to mircounts too since it uses pysam of python2 right now)

ccedmendoza: nevermind I remember that it uses smRtools...

ccedmendoza:
I don't see ainything wrong here

tools/small_read_size_histograms/size_histogram.xml
drosofff: Yeah. I still have the shift to py3 in mind. But I want to clear off smRtools module before. I'll change to py3 when I remove the smRtools dependency from size_histogram.py.
You may do the same for your current dev, since you are also squeezing smRtools ;-)

drosofff:
Thanks @ccedmendoza

PR #153, opened 2017-07-16
Chrissmallrnamap

drosofff:
@chamaelj 
Here is a major PR that includes
- clarification and simplification of the python code
- The test of the ggplot2-base r-script revealed that it is very slow (45h in a real-configuration test). I tried my best to optimize but could not find a way to improve significantly.
- I parallel preliminary test of lattice showed that it outperforms by 2-log the ggplot code. Therefore, I started to develop a lattice code. Here, the first part is done, with read map. Mainly, I worked from the code in https://github.com/ARTbio/tools-artbio/blob/master/tools/msp_sr_readmap_and_size_histograms/plot_size_readmap.r. I think it should not be too difficult to add now the plots with median and/or mean sizes (whose interest still need to be clarified as currently presented) using grid and gridextra.


drosofff:
@chamaelj I made a mistake, this PR was to merge into smallrnamap, not master yet

PR #154, opened 2017-07-17
my vision of mircounts.py

drosofff:
@ccedmendoza  we will need to fix some divergencies but it's a minor issue.
I'll start to fix the tests tomorrow (the R code needs to be updated because this tools does NOT deal with samples names (monosample outputs)


drosofff:
@ccedmendoza I think the tests will pass now. Feel free to merge in the mircounts branch when you are ready

drosofff:
ping @ccedmendoza I think the tool is ready to go to the toolshed. I think we can derive rapidly a tRF_count tool which will be super simpler and should interest @chamaelj  for her internship

PR #160, opened 2017-07-24
Yac output file label modification

ccedmendoza:
The output file label has been modified as requested in issue #7


drosofff:
@ccedmendoza better to work with clean PR. Make a branch from an up to date pulled master branch before incorporating work on yac in it !

PR #168, opened 2017-07-26
Yac accept and returns fastqsanger format

drosofff:


tools/yac_clipper/yac.xml
```diff
@@ -41,7 +41,7 @@
     <outputs>
         <data format="fasta" metadata_source="input" name="output" label="Clipping of ${input.name}">
           <change_format>
-              <when input="out_format" value="fastq" format="fastq" />
+              <when input="out_format" value="fastq" format="fastqsanger" />
```
mvdbeek: Given that yac.py does not manipulate the quality score, I think `format_source="input"` would be more correct, while having the intended effect (that fastqsanger input produces fastqsanger output, right?)

tools/yac_clipper/yac.xml
```diff
@@ -9,12 +9,12 @@
                                        --Nmode $Nmode
   </command>
     <inputs>
-        <param format="fastq" label="Source file" name="input" type="data" />
+        <param format="fastq,fastqsanger" label="Source file" name="input" type="data" />
```
mvdbeek: Since fastqsanger is a subclass of fastq this change should have no effect.

tools/yac_clipper/yac.xml
mvdbeek: That is if format_source works in a `<change_format>` tag ...

drosofff: Yes but format_source is not allowed in when statements. Any idea ?

mvdbeek: Except for fixing galaxy ... I guess your version is fine for the moment.

mvdbeek: wait, we can turn this around ... make the default format_source, and then in the when statement `<when input="out_format" value="fasta" format="fasta"`

drosofff: Right ! thanks

PR #174, opened 2017-08-23
Add date to retrieve fasta

ccedmendoza:
Modified both python and xml script so as to add a 'filter sequences per date' option


tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -56,6 +56,19 @@ def __init__(self, options, logger):
         self.count = 0
         self.webenv = ""
         self.query_key = ""
+        self.datetype = options.datetype
+        if options.reldate:
+            self.reldate = options.reldate
+        else:
```
drosofff: Why this `else` ?

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -56,6 +56,19 @@ def __init__(self, options, logger):
...
+            self.reldate = ''
+        if options.mindate:
+            self.mindate = options.mindate
+        else:
```
drosofff: why this `else` ?

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -56,6 +56,19 @@ def __init__(self, options, logger):
...
+            self.mindate = ''
+        if options.maxdate:
+            self.maxdate = options.maxdate
+        else:
```
drosofff: why this `else` ?

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -310,12 +329,27 @@ def __main__():
...
     parser.add_option('--loglevel', choices=LOG_LEVELS, default='INFO', help='logging level (default: INFO)')
     parser.add_option('-d', dest='dbname', help='database type')
     (options, args) = parser.parse_args()
     if len(args) > 0:
         parser.error('Wrong number of arguments')
-    
+    if (options.reldate and options.maxdate) or (options.reldate and options.mindate):
+        parser.error("You can't mix 'reldate' and 'maxdate', 'mindate' parameters")
+    if (options.mindate and not options.maxdate) or (options.maxdate and not options.mindate):
+        parser.error("mindate and madate must be used together")
```
drosofff: madate --> maxdate

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -310,12 +329,27 @@ def __main__():
...
+    parser.add_option('--maxdate', dest='maxdate', help='Date range used to limit a search result by the date specified by datetype. These two parameters (mindate, maxdate) must be used together to specify an arbitrary date range. The general date format is YYYY/MM/DD, and these variants are also allowed: YYYY, YYYY/MM.')
+    parser.add_option('--mindate', dest='mindate', help='Date range used to limit a search result by the date specified by datetype. These two parameters (mindate, maxdate) must be used together to specify an arbitrary date range. The general date format is YYYY/MM/DD, and these variants are also allowed: YYYY, YYYY/MM.')
     parser.add_option('--loglevel', choices=LOG_LEVELS, default='INFO', help='logging level (default: INFO)')
     parser.add_option('-d', dest='dbname', help='database type')
     (options, args) = parser.parse_args()
     if len(args) > 0:
         parser.error('Wrong number of arguments')
-    
+    if (options.reldate and options.maxdate) or (options.reldate and options.mindate):
```
drosofff: Can you enforce users not mixing parameters by manipulating the arguments ?

On a side note, the xml plays the role of keeping parameters consistant. So it may be unnecessary to deal with this

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.xml
```diff
@@ -37,6 +55,15 @@
         <output name="outfilename" ftype="fasta" file="output.fa" />
         <!--  <output name="logfile" ftype="txt" file="log.txt" />  log.txt changes with timestamp. removed to pass the  test -->
     </test>
+    <test>
+        <param name="queryString" value="9629650[gi]" />
```
drosofff: Maybe you could use another ID for another test. It's the occasion to increase the scope of the tests !

drosofff:
Also, try to lint the python code with flake8 if it's not already done (I wish to add a flake8 linting in the travis test at some point)

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
ccedmendoza: else is in case the option isn't used there is no need to check it afterwards.
In this case the url will only state &reldate=&nextoption.
If I don't use it I have to check if the option exists with an if anyway.

tools/fetch_fasta_from_ncbi/retrieve_fasta_from_NCBI.py
```diff
@@ -310,12 +329,27 @@ def __main__():
...
+    if (options.mindate and not options.maxdate) or (options.maxdate and not options.mindate):
+        parser.error("mindate and madate must be used together")
+    if options.datetype == 'publication_date':
+        options.datetype = 'pdat'
+    elif options.datetype == 'modification_date':
+        options.datetype = 'mdat'
+    elif options.datetype == 'entrez_date':
+        options.datetype = 'edat'
+    else:
+        parser.error('The date type options are [modification_date, publication_date, entrez_date]')
```
drosofff: I don't think this block (starting from the previous `if`) is necessary:

you are duplicating the job of argparse that will complain if the options are mistyped.

You can put directly 'pdat' in the argparse instance (or let 'publication_date' as options.datetype value, as you wish)

ccedmendoza: Ok, I actually did it for th options t be easier to understand

drosofff: The help field of argparse instance is your friend ;-)

drosofff: The lesser lines, the easier to understand the code

drosofff:
When the tests pass, think to:

- change to owner to `artbio`
- bump the version to 2.0.0 (this is a major release due to the creation of a new repo with a new artbio owner)

drosofff:
🥇 👍

Issue #148, opened 2017-07-08
How to separate small RNA reads that show 10nt overlap using signature.py


juulluu21:
Is there any way that after running the _signature.py_ script, I can get the reads that produce a specific overlap (for example I want to get all the reads that are overlapping by 10nt)?

Thanks!

juulluu21:
Is there anyone who has assigned to do this? Or at least, is it possible to do?
Thanks.

drosofff:
@juulluu21 with the last update of signatures tool (PR #176), it will be easier to address the need !
I would favor a separate tool such as `overlapping reads`

juulluu21:
@drosofff: thank you very much. The command I run the script is:
signature.py  input.sam  min_length   max_length    min_overlap    max_overlap  output.txt

To get the overlapping (10nt overlap) reads (all sizes), how the command would look like.
Sorry for the novice question.

drosofff:
@juulluu21 it will be something like:

`overlapping_reads.py input.sam min_length max_length min_overlap max_overlap output.fa(stq)`... but I can't really answer before coding the tool ;-)

juulluu21:
Hahaha.....

drosofff:
@juulluu21 see the PR #191 and the argparse instance in the python code for direct use of the python code. Test is welcome before pushing the tool in the toolshed !
Note that for in line python run, the bam file must be next to the index (.bai) for proper use of the pysam module.

juulluu21:
Thank you very very much. I have used the script. I used this command in mac terminal:

python overlapping_reads.py  --input  xxx.sorted.bam   --minquery  26    --maxquery  26  --mintarget  22  --maxtarget  22  --overlap 10   --output xxx.fastq

Output:
This is (below) what I got in the output file. It seems a fasta file instead of fastq. I expected that the output file would contain only 26 and 22nt reads. But, it has different sizes reads (26-31nt). Is the command that I am using is wrong?


>scaffold_1|7774354|R|26
TCCGTCTGAAGTTGATGAACCTGATG
>scaffold_1|7769043|R|26
TAATGACAAAAAATCTTGAGTGGACT
>scaffold_1|7570282|R|30
TAGGACGTAAGCTACATTAGAATGTCAACA
>scaffold_1|7570282|R|30
TAGGACGTAAGCTACATTAGAATGTCAACA
>scaffold_1|7570282|R|30
TAGGACGTAAGCTACATTAGAATGTCAACA
[...]

PR #191, opened 2017-08-29
Overlapping_read.py

drosofff:
as requested in #148 !


drosofff:
@juulluu21 I am going to merge this PR since all tests are OK. Let me know if you need clarification on the "pairable reads" (the concept of "paired reads" is not easy to handle with not uniquely mapped reads)

PR #199, opened 2017-09-06
Removed 'edat' from retrieve because it is only available for pubmed

ccedmendoza:
Since we only use the nuccore and prot databases 'edat' is useless in this case


drosofff:
What do you mean by "only available for pubmed" ?

ccedmendoza:
It is only available for the pubmed database and not the nuccore or prot databases.

drosofff:
Ah Ok !

drosofff:
Can you bump the tool to 2.0.1 ?

PR #216, opened 2017-10-03
Optimize blast unmatched

ccedmendoza:
The script had a hight memory consumption, especially when handling 120 million lines files.


drosofff:
OK 👍

PR #225, opened 2017-10-13
cleaning of october 13th 2017

drosofff:


drosofff:
there seems to be an issue with planemo but I merge this PR avec testing with planemo 0.40.1

PR #243, opened 2017-11-08
Fetch fasta

ccedmendoza:
This is the addition of the dry run option


drosofff:
Good to have a dry option. Prefiguring a multi-request tool with reset of REST between requests ;-)

PR #245, opened 2017-11-22
Fetch fasta upgrade

ccedmendoza:
Just a pull request. It doesn't work yet


ccedmendoza:
The tool failed a retrieve of 1 000 000 sequences when parallelized 4 ways. But it correctly retrieved the 1 000 000 sequences when parallelized 2 ways.
It is currently being tested for a 2 000 000 sequences retrieval while parallelized 2 ways.

ccedmendoza:
The previous test wasn't successful.
The new test is based on the fact that a epost isn't necessary as long as the request has a maximum of 200 sequences. So using the fact that the script now is able to fetch sequences using their ids we just limit the efetch to 200 ids in the loop.

ccedmendoza:
The 2 million sequence retrieval during peak hour test passed successfully.
The retrieval took 3 hours.

drosofff:
Yep, sounds good. Let's see some tests for this new mode.

drosofff:
@ccedmendoza can you put something more explicit than `Dry run to get the number of sequences?` for the label of a retrieve only unique identifiers case ?

I think also that the help section of the tool should be updated. This tool may have a large audience.

tools/fetch_fasta_from_ncbi/fetch_fasta_from_NCBI.xml
```diff
@@ -112,16 +112,20 @@
...
-This tool retrieves nucleotide/peptide sequences from the corresponding NCBI database for a given entrez query.
+This tool retrieves nucleotide/peptide sequences from the corresponding NCBI database (nuccore or protein) for a given entrez query.
 
 The tool is preset with "txid10239[orgn] NOT txid131567[orgn] AND complete NOT partial[title] NOT phage[title]" for metaVisitor use purpose
 
 See `Entrez help`_ for explanation of query formats
 
 Be sure to use the appropriate NCBI query syntax. Always use [] to specify the search fields.
 
+By checking the checkbox you can also run your query without seqence retrieval and get the number of sequences your query will fetch.
```
drosofff: sequence

tools/fetch_fasta_from_ncbi/fetch_fasta_from_NCBI.xml
```diff
@@ -112,16 +112,20 @@
...
+
 Note that the tool may fail in case of interrupted connexion with the NCBI database (see the log dataset)
 
+You can also check the retrieval progress by reading the log file in which you can see the number of sequences the query will retrieve as well as the number of squences that have been retrieved. The latter can be infered by checking the number of groups that have been retrieved (a group represents at most 200 000 sequences).
+
```
drosofff: I would just put
```
Retrieval progress is reported in the log dataset.
```

PR #257, opened 2018-01-31
fix unrecognized fastsanger datatype and bump to 2.0.2

drosofff:
There must be a better way to detect fastqsanger since it is a subclass of fastq, but this probably covers the majority of cases.
Also add a test for fastqsanger datatype.


tools/sr_bowtie/sRbowtie.xml
```diff
@@ -14,7 +14,7 @@
         #end if
         #if $input.extension == "fasta":
             #set format = "-f"
-        #elif $input.extension == "fastq":
+        #elif $input.extension == "fastq" or $input.extension == "fastqsanger":
```
mvdbeek: you're probably looking for `$input.is_of_type('fastq')` as briefly mentioned in https://docs.galaxyproject.org/en/master/dev/schema.html#id4

drosofff: Yes it is exactly what I was looking for ! Thanks @mvdbeek !

PR #260, opened 2018-02-14
added an empty blast input handling, tests and test files

ccedmendoza:


drosofff:
OK, so I am not opposed to the suppression of the "empty file" test, given what we discussed this afternoon ;-)

drosofff:
Thanks ! 👍

PR #262, opened 2018-02-20
fix input format

ccedmendoza:
If not fixed the tool can't replace fastq-to-fasta in workflow use case 2-2


drosofff:
OK, thanks for the catch

drosofff:
... but you are supposed to bump the tool version ;-)

PR #265, opened 2018-03-06
Upgrade small rna maps single display

ccedmendoza:


drosofff:
looks good. Can you comment on the R code change ?

ccedmendoza:
Sure, I just used the same way of plotting 2 plots and adapted it to a single plot.
This has the advantage of using grid and so I was able to change the ylabel based on the number of charts in the page.

drosofff:
Nice. I think I tried this in the past but did not finished. It had to be done anyway, Thanks.

PR #270, opened 2018-03-23
Small rna maps separate scales

ccedmendoza:


drosofff:
@ccedmendoza looks OK. Did you test how the outputs evolve when adding more samples (columns) ?

ccedmendoza:
not yet. I'm on it

ccedmendoza:
It's not ready yet:

- [ ] When there are more than 2 samples from the third column onward the plots are empty.
- [ ] Since the chromosome name was put on the left side the left label needs to be repositioned

drosofff:
@ccedmendoza then I suggest that you pull away a few tests from the current wrapper (they are often redundant), and replace them with tests with 4-6 samples/columns.

drosofff:
@ccedmendoza you can also add another point in your check list:

hard code a ylim in the input dataframe !

This is on my todo list, but it's maybe more agile if you are deep in the scripts ;-)

drosofff:
@ccedmendoza can you push your latest dev ? I will derive a branch of this branch for the max functionality. and maybe also, group the histogramme with a unique y axis for easier comparison of distributions (a request from Laure Teysset).

ccedmendoza:
+ The left label positioning isn't fixed yet.
+ The form was changed so as to avoid having a different number of normalization values and samples.
+ Tests' outputs haven't been changed yet

tools/small_rna_maps/small_rna_maps.xml
```diff
@@ -86,15 +87,15 @@
...
-        <param name="normalize" type="boolean" label="Is there a need for normalization?" truevalue="yes" falsevalue="no" checked="false"/>
+        <param name="normalize" type="boolean" label="Will there be normalization?" truevalue="yes" falsevalue="no" checked="false"/>
         <when value="no">
             <param name="inputs" type="data" format="bam" label="Select multiple alignments to parse" multiple="True"/>
         </when>
         <when value="yes">
             <repeat name="series" title="Normalization factors" min="1">
-                <param name="inputs" type="data" format="bam" label="Select alignment to parse" />
+                <param name="inputs" type="data" format="bam" label="Select multiple alignments to parse" />
                 <param name="normalization" type="text" label="Enter size/normalization factors. Leave blank for no normalization (default)"
```
drosofff: This is weird: if leave blank for default, it means that blank is equivalent to "1". Then in the example you put '1', which is misleading because it may suggest that '1' is different of " ".
I suggest another approach: put 1 in the form as default value.

tools/small_rna_maps/small_rna_maps.r
```diff
@@ -101,13 +102,19 @@ globalbc = function(df, global="", ...) {
...
+        ylimits=c(min,max);
+    }
     if (method == 'Counts') {
         p = xyplot(Counts~Coordinate|factor(Dataset, levels=unique(Dataset))+factor(Chromosome, levels=unique(Chromosome)),
         data=df,
         type='h',
         lwd=1.5,
         scales= list(relation="free", x=list(rot=0, cex=0.7, axs="i", tck=0.5), y=list(tick.number=4, rot=90, cex=0.7)),
-        xlab=NULL, main=NULL, ylab=NULL,
+        xlab=NULL, main=NULL, ylab=NULL, ylin=ylimits,
```
drosofff: warning: ylin instead of ylim ?

drosofff:
Thank you very much @ccedmendoza really useful work ! Hope Antoine and Laure will appreciate ;-)

PR #278, opened 2018-04-10
Small rna maps change axis label

ccedmendoza:
Changed 'Coordinates (nbre of bases)' to 'Coordinates (nbre of bases)' and removed unused test-data


drosofff:
Thanks @ccedmendoza 

> Changed 'Coordinates (nbre of bases)' to 'Coordinates (nbre of bases)'

The change is not obvious 😈

PR #284, opened 2018-05-16
Shedtoolconf parser

drosofff:
The shed_tool_conf.xml configuration file tends to become more and more complex over time and tool installations.
Thus `<section>` with same attributes id and name (identity of submenus) are sparse in  shed_tool_conf.xml which complicate the management of submenus in the Galaxy tool panel.
This tool defragments the `<section>` tags.
Thus, after cleaning, each submenu in the Galaxy tool panel have a single `<section>` in shed_tool_conf.xml whose `<tool>` children correspond to the tools hosted in this submenu.
Note that the tools also backup the original shed_tool_conf.xml before xml manipulation.


drosofff:
I will ignore the security warning from codacy for the moment

PR #285, opened 2018-05-16
Gzip wrapper : justgzip

bellenger-l:
Creation of a galaxy tool named justgzip using gzip for file compressing.
- Only take fastq files


drosofff:
Well done ! bunch of thanks !

PR #286, opened 2018-05-18
removed -k from command and added autodetection of input sequence type

michelraymond:


drosofff:
@michelraymond Although the approach works for the tool used in standalone, there is a major issue when the tool is used in a workflow:

The choice of displaying output dataset is controlled by a filters at run time. Thus, in a workflow the user experiences in any cases 4 outputs, and may well scratch his/her head before connecting one of these outputs to the next step.

I tried to find another way to select the output format. Unfortunately, I could not find something really satisfactory, in balance with letting the user choose the output format manually, which is not so terrible and does not complicate workflows either.

PR #287, opened 2018-05-22
Add gz to justgzip label

ccedmendoza:
Did this to make it easier to differentiate input and output in the history


drosofff:
@ccedmendoza looks good. thank you

PR #296, opened 2018-07-20
Add create-collection tool

ccedmendoza:
create_collection is a tool meant to be used in a workflow.
It takes multiple datasets and returns a dataset collection.

*WARNING*: The collection elements are free from metadata and have a **data** datatype so all metadata is to be set in the workflow.

### DIFF
--- a/tools/create_collection/create_collection.xml
+++ b/tools/create_collection/create_collection.xml
@@ -1,0 +1,80 @@
+<tool id="create_collection" name="Create_dataset_collection" version="0.1.0">
+  <description>of type list</description>
+  <stdio>
+      <exit_code range="1:" level="fatal" description="Tool exception" />
+  </stdio>
+  <command detect_errors="exit_code"><![CDATA[
+       mkdir outputs; cd outputs;
+       #for $input in $input_list#
+         #set $file_name =$input.name#
+         echo "$file_name">> $log;
+         ln  -s  $input "$file_name";
+       #end for#
+       ls ;
+  ]]></command>
+  <inputs>
+      <param name="list_label" type="text" label="Collection label"
+          help="Label the collection should be given." value="Collection label">
+          <validator type="length" min="1" max="50"/>
+      </param>
+      <param name="input_list" type="data" label="Input datasets"
+          multiple="true" help="Datasets to create a collection with." />
+  </inputs>
+  <outputs>
+      <collection name="output_list" type="list" label="${list_label}" format_source="input_list">
+          <discover_datasets pattern="__name__" directory="outputs" visible="false"/>
+      </collection>
+      <data format="txt" name="log" label="Create list log"/>
+  </outputs>
+  <tests>
+      <test>
+          <param name="list_label" value="Collection test"/>
+          <param name="input_list" value="file1.txt,file2.txt,diff.txt"/>
+          <output_collection name="output_list" type="list">
+              <element name="file1.txt">
+                  <assert_contents>
+                      <has_text text="Je fais souvent ce reve etrange"/>
+                  </assert_contents>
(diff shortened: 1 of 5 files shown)


tools/create_collection/create_collection.xml
```diff
@@ -0,0 +1,80 @@
...
+
+Takes any datset (as a data format).
+
+**Outputs**
+      
++ A **list** type dataset collection.
+
+      .. class:: warningmark
+      
+      *NOTE*: This tool will not keep the input datasets metadata. You should manually add it to your outputs in the workflow editor.
```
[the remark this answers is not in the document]
ccedmendoza: I managed to keep the format of the dataset but I haven't found a way to keep the reference genome.

tools/create_collection/create_collection.xml
```diff
@@ -60,26 +60,26 @@
 
 .. class:: warningmark
       
-*NOTE*: Be advised, to use the in workflows. And prioritize the use of the normal 'checkbox in dataset history' tool otherwise.
+*NOTE*: Be advised, to use this tool in workflows. And prioritize the use of the normal 'checkbox in dataset history' tool otherwise.
```
drosofff: This note is redundant with the warning in "what it does" (plus not really meaningful)

tools/create_collection/create_collection.xml
```diff
@@ -60,26 +60,26 @@
...
 .. class:: infomark
 
 **What it does**
 
-Creates a dataset collection.
+Creates a dataset collection of type list.
+
+.. class:: warningmark
+      
+      + The tool was designed to be used in a workflow. Prioritize the use of the checkbox in dataset history when creating collections when not in a workflow.
```
drosofff: @ccedmendoza can you improve the purpose of this tool, I find it hard to understand. I would put something like "This tool allows to programmatically creates a list (collection) of datasets in a Galaxy workflow"

tools/create_collection/create_collection.xml
```diff
@@ -60,26 +60,26 @@
...
+      + This tool will keep the datasets formats but won't keep the datasets reference genome.
```
drosofff: This is a bit cryptic: "won't keep the datasets reference genome" ?

tools/create_collection/create_collection.xml
```diff
@@ -60,26 +60,26 @@
...
       
 **Inputs**
 
-Takes any datset (as a data format).
+Takes any dataset (as a data format).
```
drosofff: What does mean "as a data format" ?

tools/create_collection/create_collection.xml
```diff
@@ -0,0 +1,87 @@
...
+       #for $input in $input_list#
+         #set $file_name =$input.name#
+           #if $file_name in $names#
+             #silent $names[$file_name] = $names[$file_name] + 1#
+             #set $file_name = ''.join([$file_name,'(',str($names[$file_name]-1),')'])#
+           #else#
+             #silent $names[$file_name] = 0#
+           #end if#
+         echo "$file_name" >> $log;
+         ln  -s  $input "$file_name.$input.ext.$input.dbkey";
```
mvdbeek: The symlinks will break on galaxy instances that use more than one object stores which might be on different disks.

There will however be a couple of new collection tools in the upcoming 18.09 release (https://github.com/galaxyproject/galaxy/pull/6312/files) that will allow you to create collections form datasets and datasets from collections (in workflows or otherwise) without increasing disk consumption. 

If you want to beta test them feeedback is always very welcome!

ccedmendoza:
Since a version of this tools is being implemented by default in the next version of galaxy, there is no need for this one.

PR #297, opened 2018-08-30
Remove --no-polyA which is not a valid argument for rsem-1.3.0

moskalenko:
This removes --no-polyA, which causes a rsem-prepare-reference error at tool runtime when the corresponding interface option is selected.

### DIFF
--- a/tools/rsem/macros.xml
+++ b/tools/rsem/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
   <macros>
-    <token name="@WRAPPER_VERSION@">0.8.0</token>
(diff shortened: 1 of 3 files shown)


drosofff:
Thanks for the catch @moskalenko !
Would you please bump the version of the wrapper to 0.9.0 in https://github.com/ARTbio/tools-artbio/blob/95320720112708620f39ec5843f49ca1b442ae1f/tools/rsem/macros.xml#L3 ?

moskalenko:
Version updated and commit squashed.

drosofff:
All good ! Thank you very much @moskalenko

PR #302, opened 2018-09-26
add test to check if aligned and unaligned reads files have the same …

ccedmendoza:
…format as the input


ccedmendoza:
[SrBowtie tools has an issue with outputs when input data are fastq files](https://trello.com/c/g3evKdGl/368-srbowtie-tools-has-an-issue-with-outputs-when-input-data-are-fastq-files)

drosofff:
Nice catch for the format_source ! 👍

PR #304, opened 2018-10-01
Improve small rna maps

ccedmendoza:
not done but the improvements are there


ccedmendoza:
[Improve small_rna_maps speed](https://trello.com/c/3A19FOtX/379-improve-smallrnamaps-speed)

tools/small_rna_maps/small_rna_maps.py
```diff
@@ -54,15 +54,13 @@ def create_map(self, bam_object, minsize, maxsize):
             map_dictionary[(chrom, 1, 'F')] = []
             map_dictionary[(chrom, self.chromosomes[chrom], 'F')] = []
             for read in bam_object.fetch(chrom):
-                if (read.query_alignment_length >= minsize and
-                        read.query_alignment_length <= maxsize):
```
ccedmendoza: The reason a test output was changed is that the previous version of small_rna_maps filtered not on the read length, as said in the documentation, but on the alignment length.

tools/small_rna_maps/small_rna_maps.py
```diff
@@ -190,14 +191,23 @@ def compute_coverage(self, map_dictionary, out, quality=10):
...
+                coverage = [sum(x) for x in zip(*coverage)]
+                coverage_dictionary[key] = coverage[0]
+            except TypeError:
+                if key[0] not in missing:
+                    missing[key[0]] = [key[1]-1]
+                else:
+                    missing[key[0]].append(key[1])
+        if len(missing) > 0:
+            for chrom in sorted(missing):
+                 coverage_dictionary[chrom] = [0]*(missing[chrom][1] - missing[chrom][0]+1)
```
drosofff: @ccedmendoza could you elaborate a little bit on the missing dictionary and its purpose ?

ccedmendoza: @drosofff the `key` in `for key in map_dictionary:` contains every chromosome of the reference. Which is normal since we want to output empty graphs if the chromosome wasn't mapped.
The problem is that we can't use the   `count_coverage` if there is no reads (no positions in map_dicrionary for this chromosome). So the missing dictionary is here to handle unaligned chromosomes. By assigning the value 0 to all their positions.

drosofff:
> https://github.com/ARTbio/tools-artbio/pull/304/commits/36bb8cf170bee127e886c907b439a667a82fb6c0 
Are you trying to get me lost ? If yes, it is a success 😄 
At some point, I will ask you to squash your commits, for sake of clarity

tools/small_rna_maps/small_rna_maps.py
```diff
@@ -210,7 +210,7 @@ def compute_size(self, map_dictionary, out):
         #  to track empty chromosomes
         for chrom in self.chromosomes:
             if self.bam_object.count(chrom) == 0:
-                size_dictionary[chrom]['F'][10] = 0
```
ccedmendoza: This was what added counts to the unaligned chromosomes

tools/small_rna_maps/small_rna_maps.py
```diff
@@ -212,7 +210,7 @@ def compute_size(self, map_dictionary, out):
...
+                size_dictionary[chrom] = 0
```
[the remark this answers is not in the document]
ccedmendoza: `size_dictionary[chrom]['F'][10] = 0` didn't raise the KeyError you were expecting to handle in line 248.
So I changed it to `size_dictionary[chrom] = 0` that raises a TypeError and handle it.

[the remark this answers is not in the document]
drosofff: Woah ! I don't catch the details of the reasoning, but Assigning a wrong type to a variable to raise an error to handle it sounds like.... crap....

ccedmendoza:
Sorry, I made a mistake the tool was finished here: https://github.com/ARTbio/tools-artbio/pull/304/commits/34cba44051d80008aeeb7ee20a62ba79b62b452b
The rest is just me misunderstanding and trying to fix things that aren't broken.
I will close this PR and re-open it with the fixed branch.

PR #305, opened 2018-10-05
Improve small rna maps

ccedmendoza:
Add a filter on unmapped reads and change filter from alignment length to read length


ccedmendoza:
[Improve small_rna_maps speed](https://trello.com/c/3A19FOtX/379-improve-smallrnamaps-speed)

drosofff:
@ccedmendoza can you add a NEW test (additional) that make use of the dataset 6 of the history https://toolshed.g2.bx.psu.edu/view/devteam/samtool_filter2/56c31114ad4a in addition to the already available input1.bam ? 
The idea is to test when a chromosom has alignments with one sample and no alignment at all with another sample

PR #317, opened 2018-10-26
Add tool : Filter_single_cell_expression_matrix

bellenger-l:
Add new tool that can filter cells(columns) and genes (row) from an expression matrix based on different quality controls.


tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,439 @@
...
+   </requirements>
+   <stdio>
+      <exit_code range="1:" level="fatal" description="Tool exception" />
+   </stdio>
+   <command detect_errors="exit_code"><![CDATA[ Rscript $filtering_out ]]></command>
+   <configfiles>
+      <configfile name="filtering_out">
+<![CDATA[
+## Setup R error handling to go to stderr
+options( show.error.messages=F,  error = function () { cat( geterrmessage(), file=stderr() ); q( "no", 1, F ) } )
```
drosofff: @bellenger-l can you set a filter_cells_and_genes.R script call by the wrapper ? see https://github.com/ARTbio/tools-artbio/blob/master/tools/small_rna_maps/small_rna_maps.xml#L42 for an example

drosofff:
I have no idea why the travis tests fails with that line !

https://travis-ci.org/ARTbio/tools-artbio/jobs/446770315#L585

Do you have an idea ?

ccedmendoza:
When running flake8 on blast_unmatched.py I get the same error. The reason for this is that since the last PR flake8 was upgraded from 3.5.0 to 3.6.0. In python 3.8 escaping in a string that isn't raw breaks code. I will fix this with PR: #318

tools/filter_cells_and_genes/filter_cells_and_genes.xml
bellenger-l: Ok no problem !

drosofff:
Yeah.  Yet, how this lea’s branch is connected with blast_unmarched.py ?........

ccedmendoza:
The lint.sh script is run in the tools-artbio directory so it lints all the scripts every time

drosofff:
@ARTbio/artbio-staff Sympa vos nouveaux avatars ;-) 
On another note, could you think about a more "expressive" tool name ? 
I find `Filter cells and genes` cryptic enough...

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+      <param name="file" type="data" format="txt" label="Raw counts of expression data"/>
+      <param name="file_sep" type="select" label="Input column separator">
+         <option value="tab" selected="true">Tabs</option>
+         <option value=",">Commas</option>
+         <option value=" ">Whitespaces</option>
+         <option value=".">Dots</option>
+         <option value="_">Underscores</option>
+         <option value="-">Dashes</option>
+   </param>
+   <param name="file_header" type="select" label="Consider first line of input file as header?">
```
drosofff: Test what happens if no header, and add the test in the tool tests

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+      <option value="TRUE" selected="true">Yes</option>
+      <option value="FALSE">No</option>
+   </param>
+   <conditional name="option">
+      <param name="is_mito" type="select" label="Are there mitochondrial genes in the expression data ? Do you want to remove cells that detect more than 20% of mitochondrial genes ?">
```
drosofff: 2 questions in a labels is one in excess ! 

```
label="Do you want to remove cells that detect more than 20% of mitochondrial genes ? (only applicable if mitochondrial genes are in the expression data)"
```

Moreover, this has to be tested, naturally.

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+         <output name="output_matrix" file="counts_with_mito_Seurat_filtered.tab" ftype="tabular"/>
+    </test> 
+   </tests>
+   <help>
+
+**What it does**
+
+Filter_cells_and_genes takes a tabular table of raw read counts, each column corresponding to a sample count,
+and returns (1) a PDF with QC (quality controls) plots and (2) a filtered table of counts.
+
```
drosofff: A pdf with quality controls (no need for abréviation here)... This is cryptic. You have to explain

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+      </when>
+   </conditional>
+   <conditional name="method">
+      <param name="package" type="select" label="Which package/method do you want to use ?">
+         <option value="scater" selected="true">scater</option>
+         <option value="Seurat">Seurat</option>
+      </param>
+      <when value="scater">
+         <param name="gene_filter" type="select" label="How do you want to filter out genes ?">
+            <option value="low.abundances" selected="true">low abundances genes based on library size</option>
```
drosofff: low abundances genes `relative` to library size

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+            <option value="min.cells">genes that aren't detected in a certain number of cells</option>
```
drosofff: genes that are not detected in a `fraction` of cells

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+         </param>
+         <param name="min_cells" type="integer" value="3" label="What threshold do you want to apply ?"/>
```
drosofff: Threshold is not meaningful in this context. 100 % sure that 95% users will guess instead of understand. On 50% of these will guess bad.
Plus, I am lost: the same threshold parameter is used for `not in a fraction of cells` and `low abundances genes relative to library size`. Does that mean that this is a "contextual parameter" ? I yes, this is not a good idea at all.

bellenger-l: Ok it was a contextual parameter, used by two different functions, I will modify that part dans separate this parameter in two.

bellenger-l:
How about filter_sc_expression_matrix ? Maybe a bit too long ?

tools/filter_cells_and_genes/filter_cells_and_genes.xml
bellenger-l: Fix Rscript and add test for this issue

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -63,11 +68,18 @@ Rscript $__tool_directory__/filter_cells_and_genes.R
...
-            <option value="min.cells">genes that aren't detected in a fraction of cells</option>
-         </param>
-         <param name="min_cells" type="integer" value="3" label="What threshold do you want to apply ?"/>
+         <conditional name="feature_removing">
+            <param name="gene_filter" type="select" label="How do you want to filter out genes ?">
+               <option value="low.abundances" selected="true">low abundances genes relative to library size</option>
+               <option value="min.cells">genes that aren't detected in a fraction of cells</option>
+            </param>
+            <when value="low.abundances">
+               <param name="min_counts" type="integer" value="1" label="Keep all genes with an average count superior to:"/>
```
drosofff: Keep all genes with an average count higher than:

By the way, the default parameter value is 1. does that mean that under the default setting, only genes with mean count equal or higher than `2` are kept ?

drosofff: OK, thanks. Thus the label `Keep all genes with an average count superior to:` was misleading too right ?

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
...
+    </test> 
+   </tests>
+   <help>
+
+**What it does**
+
+Filter_cells_and_genes takes a tabular table of raw read counts, each column corresponding to a cell and each row corresponding to a
+gene, and returns (1) a PDF with quality controls plots (like the distribution of library sizes or expressed genes by cell, frequency of 
+expression or the most expressed genes) and (2) a filtered table of counts.
+
```
drosofff: (1) a PDF with quality controls plots (see explanations below) and (2) a filtered table of counts.

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
...
+You can filter out low quality cells from your raw expression table :
+
+- Based on expression level of mitochondrial genes :
```
drosofff: Give a short explanation of mitochondrial RNA as a quality marker

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
...
+
+    - Using R package scater: remove cells that detect mitochondrial genes 3 times higher than the median of all cells
```
drosofff: Be careful with the term "mitochondrial genes" it may be misleading. I understand that you speak of genes that are encoded by the mitochondrial genome. But some others could understand that this include gene encoding mitochondrial protein, which is not the case here, right ?
"mitochondrial genes (i.e. genes mapped to the mitochondrial chromosome)" would be more explicit

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
...
+
+
+**Inputs**
+
+The following table gives an example of the tabular input that has to be provided to the tool:
+
+
+<![CDATA[
+
+AGTCAGACTAGAGAAGTCCAGATTTCAC	AGTCTACTTGCATG	AGTGTTCTATAAGG	AGTGTTCTCACTTT	AGTTCTTGAAGCCT	ATAAACACAGTGCT	ATAACCCTGTTGGT	ATAATCGAGCTGAT	ATACCACTCGTACA
```
drosofff: What the hell is this ? ;-)

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
...
+
+MT-ND2	4	4	15	2	0	1	3	9	0	8  
+
+MT-CO1	11	3	52	10	9	14	11	9	9	22  
+
+MT-CO2	4	5	35	8	5	2	7	10	2	7  
+
+MT-ATP8	0	0	0	0	0	0	0	0	0	0  
+
+
```
drosofff: ** Outputs **

And here we are. Sorry Lea but you will  have to explain every possible output, especially the pdf. This is the (nice) job: explaining the outputs is just *Science* - and make all difference with bad "have to guess" tools !

tools/filter_cells_and_genes/filter_cells_and_genes.xml
bellenger-l: This is how I understand from researches I made, it's appear that it corresponds to genes mapped to the chrM. In all tools I used, they use the term of "mitochondrial genes"

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
...
+    - Remove all genes that aren't detected in at least n cells
+    - Filter out genes based on its average count (which is adjust for library size between cells). It is more stringent.
+
+- Using R package Seurat : Remove all genes that aren't detected in at least n cells.
+
+
+**Inputs**
+
+The following table gives an example of the tabular input that has to be provided to the tool:
```
drosofff: Note that the first line show cell identifiers and table report the expression of xx gene (aka xx lines).

(be nice with your users)

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,270 @@
+<tool id="filter_cells_and_genes" name="Filter out cells and genes" version="0.1.0">
```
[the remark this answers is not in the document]
drosofff: Bump a bit your version please (should reflect the work on the code even if it is not commited

tools/filter_cells_and_genes/filter_cells_and_genes.xml
drosofff: Ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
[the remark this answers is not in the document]
drosofff: Ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
drosofff: Ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
```diff
@@ -0,0 +1,246 @@
...
+         <param name="min_cells" value="3"/>
+         <output name="output_plot" file="counts_with_mito_Seurat_filter_plots.pdf" ftype="pdf"/>
+         <output name="output_matrix" file="counts_with_mito_Seurat_filtered.tab" ftype="tabular"/>
+    </test> 
+   </tests>
+   <help>
+
+**What it does**
+
+Filter_cells_and_genes takes a tabular table of raw read counts, each column corresponding to a sample count,
```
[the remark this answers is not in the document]
drosofff: Ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
drosofff: Ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
drosofff: ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
drosofff: ping @bellenger-l

tools/filter_cells_and_genes/filter_cells_and_genes.xml
drosofff: @bellenger-l The value of a tool often stands in its documentation

tools/filter_single_cell_expression_matrix/.shed.yml
```diff
@@ -0,0 +1,10 @@
+# .shed.yml supporting automatic pushes.
+owner: artbio
+name: filter_single_cell_expression_matrix
+description: Filter out low quality cells and genes from raw scRNAseq expression matrix 
+categories:
```
drosofff: the 
```
long_description: |
  Generates hit count lists from bam alignments.
```
tag (here an example is given) has vanished here ! This is not good.

PR #318, opened 2018-10-29
set regex in raw string

ccedmendoza:


ccedmendoza:
Merging the PR so as to let PR #317 pass the flake8 linting

PR #319, opened 2018-11-07
invoice refence parsing didn't work with one of the new html files, ref parsing is now done outside the pandas parsed tables

michelraymond:


michelraymond:
[Logiciel de facturation](https://trello.com/c/yEAtQlSp/354-logiciel-de-facturation)

michelraymond:
Also btw, do you know which file in plastisipi corresponds to the galaxy sniffer? It still needs to be edited to solve the `<meta>` tag bug. It'd be great to have that solved for tomorrow's meeting :)

drosofff:
@michelraymond this is this file https://github.com/galaxyproject/galaxy/blob/release_18.05/lib/galaxy/util/checkers.py whose path is the same except the root part. Thus something as /home/galaxy/galaxy-dist/lib/galaxy/util/checkers.py

However, no need to say that changing the code in place is completely prohibited. Work on a Galaxy fresh instance in a VM or whatever

PR #335, opened 2018-12-24
small_rna_maps switches to input multiple to be able to consume data collections

drosofff:
I have not been able to find a test syntax that explicitly tests a data collection of bam files.


ccedmendoza:
In the concatenation tool PR there is a data collection test example, it's fastq but should work for bam files

drosofff:
@ccedmendoza yes I saw, but I could not make it. probably because my "inputs" is not declared as collection since it can also be used in a multiple input files context. does not mind for the moment...

Happy Christmas, and good night !

PR #360, opened 2019-05-27
[WIP] Add union/intersection of cutoff option (Filter cells)

bellenger-l:
J'ai commencé à retravailler le premier wrapper filter_cells


bellenger-l:
Close PR because of duplicated work from #361

PR #361, opened 2019-05-27
Add option of union/intersection of cutoffs in gsc-filter_cells

drosofff:
sorry @bellenger-l I scooped you...


drosofff:
bad target branch.. I close the PR

PR #400, opened 2019-06-25
Fix error color with two levels - High_dim_visu

bellenger-l:


drosofff:
@bellenger-l  The fix works, I have added the corresponding test and will merge when the travis test passes

Issue #401, opened 2019-06-26
Improvement gsc_high_dim_visu


bellenger-l:
- allow numeric factor for cell coloration based on : 
    - a score thanks to a two columns table
    - gene expression by looking directly in the expression matrix (be careful, if raw count matrix maybe log transformed the data)
- plot PC1 vs PC2 - PC1 vs PC3 - PC2 vs PC3
- add axis labels for t-SNE
- remove unnecessary parameter : normalize 
- output cluster table
- add clusterR utility to get external validation of clustering (only if the factor table is given & is not numerical)

--> Done : #405

bellenger-l:
Add UMAP visualisation ?

drosofff:
Sure, but in a new issue. In CI, what is done is done

bellenger-l:
Ok !

PR #405, opened 2019-06-28
[WIP] Update gsc_high_dimensions_visualisation

bellenger-l:
Following #401, here is my contribution : 

*DONE*
- Add numerical factor
- remove unnecessary parameter : normalize [re-integrate this option with R 3.5 upgrade]
- add axis labels for t-SNE
- add clusterR utility to get external validation of clustering (only if the factor table is given & is not numerical)
- output cluster table 
- add PC to plot parameter

*To Do*
- ...


tools/gsc_high_dimensions_visualisation/high_dim_visu.R
```diff
@@ -228,20 +222,24 @@ if (opt$factor != '') {
...
-  factorColors <-
-    with(contrasting_factor,
-         data.frame(factor = levels(contrasting_factor$factor),
-                    color = I(colors_labels)
-         )
-    )
-  factor_cols <- factorColors$color[match(contrasting_factor$factor,
+  if(is.numeric(contrasting_factor$factor)){
+    factor_cols <- brewer.pal(n = 9, name = "Blues")[contrasting_factor$factor]
+  }else {
```
drosofff: With a space is more readable

tools/gsc_high_dimensions_visualisation/high_dim_visu.R
```diff
@@ -355,11 +359,13 @@ plot(res.hcpc, choice="map")
 
 # user contrasts on the pca
 if (opt$factor != '') {
+  if(is.factor(contrasting_factor$factor)) {
     plot(pca, label="none", habillage="ind", col.hab=factor_cols)
     legend(x = 'topright', 
        legend = as.character(factorColors$factor),
        col = factorColors$color, pch = 16, bty = 'n', xjust = 1, cex=0.7)
-    }
+  }
```
drosofff: I think an else statement has been forgotten here from the beginning... unless we decide to drop completely the PCA in hcpc, which is/should be a duplication of the PCA option

bellenger-l: I cannot find how to add a legend for numerical vector for `plot`, there is a package `auto-image` with a function `legend.scale` but there is no conda package.
Unless we use ggplot for the PCA and HCPC plots.

drosofff: If we limit the issue to _PCA_ plots I think using ggplot is perfectly doable !

tools/gsc_high_dimensions_visualisation/high_dim_visu.R
```diff
@@ -272,19 +269,24 @@ if (opt$visu_choice == 'tSNE') {
     ggplot(embedding, aes(x=V1, y=V2)) +
       geom_point(size=1, color='deepskyblue4') +
       gg_legend +
-      xlab("") +
-      ylab("") +
+      xlab("t-SNE 1") +
```
drosofff: honestly, the added value of this legend is limited. Would it be possible to add useful information ?

tools/gsc_high_dimensions_visualisation/high_dim_visu.R
drosofff: Otherwise, probably some interesting material here https://stackoverflow.com/questions/13355176/gradient-legend-in-base

drosofff: and also https://aurelienmadouasse.wordpress.com/2012/01/13/legend-for-a-continuous-color-scale-in-r/

drosofff: and for ggplot visualisation of pca: https://gist.github.com/benmarwick/2139672.

drosofff: as well as http://www.sthda.com/french/articles/38-methodes-des-composantes-principales-dans-r-guide-pratique/73-acp-analyse-en-composantes-principales-avec-r-l-essentiel/

bellenger-l: Merci, je regarde tout ça

tools/gsc_high_dimensions_visualisation/high_dim_visu.xml
```diff
@@ -1,12 +1,12 @@
...
-        <requirement type="package" version="1.3.2=r3.3.2_0">r-optparse</requirement>
-        <requirement type="package" version="1.39=r3.3.2_0">r-factominer</requirement>
-        <requirement type="package" version="1.0.5=r3.3.2_0">r-factoextra</requirement>
-        <requirement type="package" version="0.13=r3.3.2_0">r-rtsne</requirement>
-        <requirement type="package" version="2.2.1=r3.3.2_0">r-ggplot2</requirement>
-        <requirement type="package" version="0.4.1=r3.3.2_0">r-ggfortify</requirement>
+        <requirement type="package" version="1.6.2=r35h6115d3f_0">r-optparse</requirement>
+        <requirement type="package" version="1.41=r351h6115d3f_1">r-factominer</requirement>
+        <requirement type="package" version="1.0.5=r351_1">r-factoextra</requirement>
+        <requirement type="package" version="0.15=r351he1b5a44_0">r-rtsne</requirement>
```
drosofff: @bellenger-l note that with this new r-rtsne version, the option `normalize` seems to be back !!!!

drosofff:
@bellenger-l I have to feeling that you are maybe missing the method to evolve tests when code is changing. The current failure is just due to a lack of test file for the added test

drosofff:
So the failure in testing is related to a story of number of digits / precision of floating number that vary between the local and remote testing environments... I cannot say more currently. Investigating.

drosofff:
the only test that does not pass is the tsne with changes in parameters perplexity, and exaggeration... maybe a clue...

bellenger-l:
Ici j'ai également une erreur sur mon planemo test en local mais il a l'air de passer les tests Travis...

tools/gsc_high_dimensions_visualisation/high_dim_visu.xml
bellenger-l: Done !

tools/gsc_high_dimensions_visualisation/high_dim_visu.R
bellenger-l: Rtsne don't give any additional information about dimensions like PCA. I always give axis labels this is why I changed to t-SNE 1 and t-SNE 2 (or Dim 1/Dim 2). If this is not relevant enough, I can get back to empty labels.

bellenger-l:
@drosofff I may have the same issue than before. My tests succeed to pass locally with planemo but the last test with Travis failed. So I add the sim size comparison for table coordinate of the last test. Now it pass both tests.

All clear for me.

PR #406, opened 2019-06-30
Update mannwhitney_de.xml

drosofff:
making factor input compatible with any number of column (only 1 and 2 are used)


drosofff:
I probably overlooked the non-existing problem. Yet this pr adds a test for a case of factor file containing more than the 2-column minimal stuff (cell id and factor levels)

PR #407, opened 2019-07-01
Make it work for use case 1-3

ccedmendoza:
Add condition so when input file is handled even if it doesn't have an extension.


ccedmendoza:
Should be the last step for all workflows to run smoothly

drosofff:
Cool. for your information, I think the last implemented workflow features should be very useful to Metavisitor ! https://training.galaxyproject.org/training-material/topics/galaxy-ui/tutorials/workflow-parameters/tutorial.html

ccedmendoza:
@drosofff , could you push the tool to the toolshed, please?

PR #414, opened 2019-07-12
fix many bugs in gsc_center_scale, previous version wrong

drosofff:
Future versions will further clarify the relationship between parameters `center` and `scale`, which is not 100% clear yet


drosofff:
:bulb:

PR #423, opened 2019-08-31
improve HCPC in gsc_high_dimension_visualisation

drosofff:
- clarify parameter usage (that is not explained in detail in the package)
- parameter names more consistent with the reduction methods
- implement cluster description by most contributing variables
- remove the useless `requiredPackages` R directive that produces a "false" bug
- adapt tests to the new hcpc functionality


drosofff:
🎉

PR #426, opened 2019-09-09
Normalization tool using scran package

bellenger-l:
New module for single cell normalization using the R package scran


tools/scran-normalize/.shed.yml
```diff
@@ -0,0 +1,12 @@
+# .shed.yml supporting automatic pushes.
+owner: artbio
+name: scran-normalize
```
drosofff: underscore (no dash) in the repo name and in the tool name

drosofff: + for homogeneity, we should prefix the tool name with `gsc_`

tools/scran-normalize/.shed.yml
```diff
@@ -0,0 +1,12 @@
...
+description: Normalize raw counts using scran  
+long_description: |
+  Normalizes raw counts expression matrix using deconvolution size factors
+categories:
+  - Transcriptomics
+homepage_url: http://artbio.fr
+remote_repository_url: https://github.com/ARTbio/tools-artbio/tree/master/tools/scran-normalize
```
drosofff: scran_normalize

tools/scran-normalize/scran-normalize.R
```diff
@@ -0,0 +1,64 @@
...
+
+## Import data as a SingleCellExperiment object
+sce <- SingleCellExperiment(list(counts=as.matrix(data)))
+
+## Compute sum factors
+sce <- computeSumFactors(sce)
+
+sce <- normalize(sce)
+
+logcounts <- data.frame(genes = rownames(sce), logcounts(sce), check.names = F)
```
drosofff: at final, what is the unit for the results of scran normalization ? cpm ? tpm ? something else I guess ?

drosofff:
@bellenger-l the test was failing for a problem of floating variables. Is it acceptable with a round at 5 digits ? This can be increased I think.

drosofff:
@bellenger-l can you add a test with no clustering please ?

bellenger-l:
@drosofff the first test is a no clustering test

drosofff:
@bellenger-l Are you sure ???
If so the parameter `param name="cluster"` has a misleading name (Value "yes" for both tests)

bellenger-l:
@drosofff yes I am, there is 3 tests : 
- first : defaults param (no clustering)
```
        <test>
            <param name="input" value="counts.tab" ftype="tabular"/>
            <output name="output" file="logcounts.tab" ftype="tabular"/>
        </test>
```
- second : cluster cell with igraph method
- third : cluster cell with hclust method

tools/scran-normalize/scran-normalize.R
drosofff: @bellenger-l any follow up on this ?

bellenger-l: @drosofff definitively not TPM because it doesn't take into account the gene length.
It creates a reference : so either the average of all transcriptomes (no clustering step) or if we choose a clustering step the average of each cluster. Then it pools cells and then sum their expression profiles. The size factor is described as the median ration between the count sums and the average across all genes. Finally it constructs a linear distribution (deconvolution method, Lun et al., 2016) of size factors by taking multiple pools of cells.

drosofff: OK, I see. So we could say that these are "scran normalized count values". Can you say me again whether the tool log-transforms the data (automatically or optionally). This is important to clearly states what it is for downstream analyses

bellenger-l: The normalize function from the scater package return by default the log transformed values.

bellenger-l: But we can choose to not log transformed the values

drosofff: good, but explain this not only to me. In the help section please. This is an essential piece of information to understand and to avoid huge bugs downstream of this tools

drosofff:
Shall we merge this / push to the toolshed ?

bellenger-l:
@drosofff if it's good for you, it's good for me

drosofff:
🎉

PR #435, opened 2019-10-25
Integration of manta wrapper

lgpdv:
Wrapper of 'manta' variant caller, for its integration into Galaxy. It has been tested with diverse data and works fine. However, the 'planemo t' command produces a bug and the test fails.


drosofff:
nice @lgpdv can you add a .shed.yml file in the tool repo, similar to https://github.com/ARTbio/tools-artbio/blob/master/tools/lumpy_sv/.shed.yml ?

drosofff:
@lgpdv unzip your test-data and code functional tests  using these data

lgpdv:
> @lgpdv unzip your test-data and code functional tests using these data

> @drosofff The problem is that zipped files are larger than the limit size of GitHub (manta requires at least 1 bam file and a (slice of a) reference genome).

drosofff:
Ba....
First of all, I am almost sure that you can have gzipped data in test-data (provided you inform the test of this status)... BUT you cannot have gzipped test-data directory (the environnement is evaluated before the tool runtime, so making the tool unzipping the test-data cannot make it).

Secondly, is it so complicated to shorten you reference ? so that the test applies only on this artificial test genome ? We are constantly doing this.

drosofff:
@lgpdv Hi Leonardo,
Do not focus to much on the test-data size:
for the moment, the tool is not working anyway.
at this stage, `planemo serve .` is still your favorite friend. (maybe a good idea if I give you a brief training on the development method at this stage of the tool, to save you time)

For your information, the current issue is

```
/tmp/tmpem0nud0v/job_working_directory/000/5/tool_script.sh: line 41: 0.7.19: command not found
Usage: configManta.py [options]

configManta.py: error: Run directory already contains workflow script file '/tmp/tmpem0nud0v/galaxy-dev/MantaWorkflow/runWorkflow.py'. Each analysis must be configured in a separate directory.
```

which suggest that the tool is implementing twice the  `runWorkflow.py file`

lgpdv:
@drosofff The bug is related with not overwriting the output. I've added a simple solution (should be more elegant ones) to the code. The training will be useful of course !

drosofff:
@lgpdv you seem stuck... did you see this https://travis-ci.org/ARTbio/tools-artbio/jobs/603452238#L1437 ?

lgpdv:
@drosofff I have tried some approaches for solving the size limit, but unsuccessfully. I will test other possibilities.

drosofff:
@lgpdv your tools does not produce outputs by default, which is a problem in your test.

lgpdv:
@drosofff Thank you very much for the hints! I will check about the desired outputs inside the test folder and I will try again the test commands.

lgpdv:
@drosofff I have modified the tests accordingly. However, some bugs are still arising...

drosofff:
@lgpdv see #436 for fixing your issues

drosofff:
@lgpdv OK... https://travis-ci.org/ARTbio/tools-artbio/jobs/608249553#L2256
... So, the next fix will be to manage memory so that the test passes on the "small" travis VMs

drosofff:
.... which hopefully should be done with https://github.com/ARTbio/tools-artbio/pull/436/commits/5c68330b6e013f125f98436bc6dae31328954b45

lgpdv:
@drosofff Thanks a lot ! You're right, seems to be the "callMemMb" parameter the responsible one for this bug.

drosofff:
@lgpdv yep, this is fixed in #436 . Note that it is not really a bug. Just tuning...

lgpdv:
Now this bug is over, but still other one has arisen...

drosofff:
> Now this bug is over, but still other one has arisen...

Not on my side 😃 . Diff and Copy are your best friends now.

For me the next step is to improve the tests: They are not stringent enough: an output with no called variant is not appropriate.
Could you enrich a bit your test dataset ?

Then the options have to be more extensively tested.

I also think that the user interface can avoid some unnecessary clics. and that the labels of the output datasets are ugggggly and scaring.

tools/manta/manta.xml
```diff
@@ -1,292 +1,293 @@
...
+		#if str( $set_configuration.set_configuration_switch ) == "Custom_config_file":
+		#set config_file = $set_configuration.CustomConfigFile
+		#else if str( $set_configuration.set_configuration_switch ) == "Customized":
+		#set config_file = config_file_custom
+		#end if
+
+		--config=${config_file}
+
+		#if str( $bam_input.bam_input_selector ) == "not_tumor_bam":
+		--bam=$input_normal
```
drosofff: The indentation is not correct here, 4 spaces to the right following the python convention

lgpdv: Ok, I will correct it.

lgpdv:
>For me the next step is to improve the tests: They are not stringent enough: an output with no called variant is not appropriate.

I've checked that obtained VCF has 6 variants..

>Could you enrich a bit your test dataset ?

I will try, but keeping in mind the 25 Mb size limit for GitHub...

drosofff:
This is easy: you add your own reads. Another advantage to this testing
approach is that you know the absolute truth since you've created it !

Christophe Antoniewski

ARTbio <http://artbio.fr/> - Tel +33 1 44 2*7 70 05*
Mobile +33 6 68 60 51 50
https://twitter.com/ARTbio_IBPS

lgpdv:
Ok, I will do like this therefore.

PR #439, opened 2019-12-13
add sigmut_wrapper

lgpdv:
Updated version of SigMatrixGeneratorProfiler


tools/sigmut/sigmut_macros.xml
```diff
@@ -0,0 +1,28 @@
...
+    <token name="@pipefail@"><![CDATA[set -o | grep -q pipefail && set -o pipefail;]]></token>
+
+    <xml name="requirements">
+        <requirements>
+		<requirement type="package" version="1.17.5">numpy</requirement>
+		<requirement type="package" version="1.3.1">scipy</requirement>
+		<requirement type="package" version="0.25.3">pandas</requirement>
+		<requirement type="package" version="0.10.1">statsmodels</requirement>
+		<requirement type="package" version="3.1.2">matplotlib</requirement>
+		<requirement type="package" version="1.0">sigmut</requirement>
```
drosofff: you can try to remove all requirements except sigmut. I think they are already included in sigmut.

lgpdv: Ok, I will do this and push PR.

drosofff:
@lgpdv can you pull the master branch in your branch ? this will skip the test and validate the PR.
I still have some suggestions after that

tools/sigmut/sigmut_macros.xml
```diff
@@ -0,0 +1,23 @@
+<macros>
+      
+    <token name="@VERSION@">11.19</token>
```
drosofff: I don't understand this VERSION variable. What is it ?

tools/sigmut/sigmut.xml
```diff
@@ -0,0 +1,495 @@
+﻿<tool id="SigProfiler" name="SigProfiler" version="1.0">
```
drosofff: Why not linking this `version` to the `@VERSION@` in macros ?

tools/sigmut/sigmut.xml
```diff
@@ -0,0 +1,495 @@
...
+	<outputs>
+		<data format="txt" name="logref" label="Log file: Install a Reference Genome">
+			<filter>set_analysis['choices'] == 'install_genome'</filter>
+		</data>
+
+                <data format="txt" name="logsmt" label="Log file: Calculate Mutational Signatures">
+                        <filter>set_analysis['choices'] == 'get_sigmut'</filter>
+                </data>
+
+		<data format="pdf" name="SBS6" label="--> Plot SBS 6 Sig. Mut.">
```
drosofff: why these weird names ? I there a biological or genomic or statistic implicit sense ?

tools/sigmut/sigmut.xml
```diff
@@ -0,0 +1,495 @@
...
+		<output name="DBS78" file="DBS_78.pdf" lines_diff="5"/>
+		<output name="DBS186" file="DBS_186.pdf" lines_diff="5"/>
+
+		</test>
+
+
+		<test>
+                        <conditional name="set_analysis">
+                                <param name="choices" value="get_sigmut"/>
+                                <param name="refgendat" ftype="fasta" value="GRCh37"/>
```
drosofff: To clarify. the Reference data are installed/downloaded when the conda package is installed ? or when the tool is running ?

tools/sigmut/sigmut.xml
```diff
@@ -0,0 +1,495 @@
...
+		--chrom			Create the matrices on a per chromosome basis
+
+		--plot			Generate the plots for each context
+
+		--tsb			Performs a transcriptional strand bias test for the
+					24, 384, and 6144 contexts
+
+		--gs			Performs a gene strand bias test
+
+		For further info see: https://github.com/AlexandrovLab/SigProfilerMatrixGenerator
```
drosofff: This tool is important. It deserves a more detailed and pedagogical help section, since it is targeted to clinician researcher.

In addition, I definitely think that the tool should be associated with the description of a real test case in a compagnon pdf with a url in the help section.

lgpdv:
@drosofff I've done what you've suggested. Now, the PR is ready to be merged. Should I do this?

tools/sigmut/sigmut_macros.xml
```diff
@@ -0,0 +1,23 @@
+<macros>
+      
+    <token name="@VERSION@">1.0</token>
```
drosofff: ... and at this stage, should not be 1.x but 0.9.0 to indicate that it is a beta version

tools/sigmut/sigmut_macros.xml
lgpdv: At the beginning I used as version ID the month + year of development. In this case, 11.19 means that this version was created in November 2019.

drosofff: this is not best practice for version values at all...

tools/sigmut/sigmut.xml
lgpdv: SBS : Single Base substitutions. The number is related to the amount of possible mutations.
The most basic classification catalogues SBSs into six distinct categories, including: C:G > A:T, C:G > G:C, C:G > T:A, T:A > A:T, T:A > C:G, and T:A > G:C. In addition, the SBS-6 classification can be further expanded by taking into account the base-pairs immediately adjacent 5′ and 3′ to the somatic mutation.

tools/sigmut/sigmut.xml
lgpdv: The reference genome(s) is(are) downloaded before calculating the mutational signatures. It is a mandatory step, because without these files the tool can not be run any further.
However, we can skip this step/option by including already downloaded files, but then the conda package will be extremely large in size. In addition, we will have to update the package very frequently for newer versions of these genome files.

tools/sigmut/sigmut.xml
lgpdv: Ok, I will modify the help section for a better explanation. Afterwards, I will work in a real example that can be reached by a URL link.

tools/sigmut/sigmut_macros.xml
lgpdv: Yes, that's OK.

tools/sigmut/sigmut_macros.xml
lgpdv: No, I guess not. Therefore, it will be '0.9.0' from now on.

tools/sigmut/sigmut.xml
drosofff: OK, I see. Thus my guess is that the failing is travis test is a time-out during this particular step, not during the conda package install.

This is actually a good news :  
Would it be possible to create a distant repo for a very slight genome prepared in silico ? The benefit from this not only being actual automated testing in travis (to which I am strongly attached as you know) but also to mastering (++ autonomy) the format of reference for the alexandrov et al tool !

tools/sigmut/sigmut.xml
drosofff: The explanation on the file names that you provided is good to integrate in the help section !

tools/sigmut/sigmut.xml
lgpdv: Yes, that's totally right.

tools/sigmut/sigmut.xml
lgpdv: Unfortunately, the reference genomes are provided by the authors as already compiled files. The script downloads the information from the following link:
ftp://ngs.sanger.ac.uk/scratch/project/mutographs/SigProf/
However, there should be a super complex way of processing a particular genome, that includes not only the fasta file(s), but also annotating all gene coordinates. I am checking this possibility, but not easy at all.

tools/sigmut/sigmut.xml
drosofff: So, go ahead !
```
<tool id="SigProfiler" name="SigProfiler" version="@VERSION@">
```
or something like that.

tools/sigmut/sigmut.xml
drosofff: But in the meantime, ftp://ngs.sanger.ac.uk/scratch/project/mutographs/SigProf/c_elegans.tar.gz is exactly what we need ! then we just have to find a test with c. elegans datasets, and I predict that the test will rock !

tools/sigmut/sigmut.xml
drosofff: Ping @lgpdv

lgpdv: Yes, that's OK. All these changes are already done, but before running the PR I would like to finish the lighter genome (c elegans) approach. It takes a lot of time because I have detected some problems with the authors' code. Locally, I have solved them and I am able to obtain the mut. sign. But, the bioconda package has to corrected...

lgpdv: I mean:
the bioconda package has to be corrected...

drosofff:
Le ven. 6 mars 2020 à 13:54, lgpdv <notifications@github.com> a écrit :

> *@lgpdv* commented on this pull request.
>
> Yes this is exactly why automated tests are good
>
-- 

Christophe Antoniewski

ARTbio <http://artbio.fr/> - Tel +33 1 44 2*7 70 05*
Mobile +33 6 68 60 51 50
https://twitter.com/ARTbio_IBPS

lgpdv:
Yes, I completely agree. Anyway, quite surprising to find code errors in an already published tool.

drosofff:
> Yes, I completely agree. Anyway, quite surprising to find code errors in an already published tool.

I am not surprised at all.

PR #455, opened 2020-05-12
Update output files from Manta wrapper.

lgpdv:
The output files located inside the 'test-data' folder has been updated.


drosofff:
Can you bump the wrapper version too ? (1.6.1 I think but I am not sure). This is important for the toolshed

lgpdv:
> 
> 
> Can you bump the wrapper version too ? (1.6.1 I think but I am not sure). This is important for the toolshed

I've just added this information.

PR #460, opened 2020-05-13
Manta: clean the mess with the .loc and .loc.sample files

drosofff:
Put them in the proper directories with the proper names


drosofff:
Finally, I add to upload manually the repo using `planemo shed_create`

PR #501, opened 2020-10-29
Fixing signature score tool

bellenger-l:
There was an error in the Rcode, two tables contained covariance matrix instead of one covariance and one correlation matrices.


bellenger-l:
@drosofff le test est passé, j'attends juste ton retour sur ce micro fix et d'accepter la PR.

drosofff:
Thanks @bellenger-l 👍

drosofff (approved):
👍

Issue #520, opened 2021-05-04
Planemo tests IFB VM


drosofff:
There were problems with 53 test(s) - out of 227 test(s) executed. See /mnt/mydatalocal/chris/tools

- [x] get_fasta_reference (Test #1): failed
- [x] lumpy_smoove (Test #1): failed
- [x] lumpy_smoove (Test #2): failed
- [x] lumpy_smoove (Test #3): failed
- [x] lumpy_smoove (Test #4): failed
- [x] manta (Test #1): failed
- [x] manta (Test #2): failed
- [x] manta (Test #3): failed
- [x] manta (Test #4): passed
- [x] manta (Test #5): passed
- [x] mircounts (Test #1): failed
- [x] mircounts (Test #2): failed
- [x] mircounts (Test #3): failed
- [x] mircounts (Test #4): failed
- [x] mircounts (Test #5): failed
- [x] mircounts (Test #6): failed
- [x] patser-v3e (Test #1): failed
- [ ] repenrich (Test #1): failed
- [ ] repenrich (Test #2): failed
- [ ] sashimi_plot (Test #1): failed
- [ ] sashimi_plot (Test #2): passed
- [ ] sashimi_plot (Test #3): passed
- [x] small_rna_clusters (Test #1): failed
- [x] small_rna_maps (Test #2..8): failed

drosofff:
I close the issue and will raise 2 new issues for
- rsem
- repenrich
- sashimi_plot (most probably)

PR #524, opened 2021-05-10
Finishing tool refresh (#520)

drosofff:
- [x] manta
- [ ] rsem


drosofff:
I leave this PR because I need to isolate the upgrade of rsem which will be "drastic"

PR #633, opened 2023-10-11
update cap3 tool

drosofff:
bump cap3 tool version

fix xml


drosofff:
let's restart this from scratch

PR #664, opened 2024-02-09
styleR

drosofff:
poke @bellenger-l and @Ning-L, this is for you.

Unfortunately not as verbose as lintR (pure tidyverse style !), but combining both tool is probably the winner combo for clean R code !

I would say to use first styler then lintr. For some already almost clean code, styler may be enough.

Note that I have the impression that both tools have different "indentation defaults" (2 for lintr and 4 for styler). To be checked but in any case this can be adjusted to you preference (or the one of tidyverse :wink:)

### DIFF
--- a/scripts/styleR/styler.R
+++ b/scripts/styleR/styler.R
@@ -1,0 +1,36 @@
+#!/usr/bin/env Rscript
+
+library("argparse")
+library("styler")
+
+parser <- ArgumentParser(description = "Call styler")
+parser$add_argument("dir",
+    metavar = "DIR", type = "character",
+    help = "File to parse"
+)
+parser$add_argument("--dry",
+    choices = c("off", "on"), default = "on"
+)
+args <- parser$parse_args()
+
+file_info <- file.info(args$dir)
+is_directory <- file_info$isdir
+
+if (is_directory) {
+    captured_output <- capture.output({
+        result <- style_dir(args$dir, indent_by = 4, dry = args$dry, recursive = TRUE)
+    })
+} else {
+    captured_output <- capture.output({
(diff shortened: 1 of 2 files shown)


Ning-L:
I've only used `lintr` on my code and I'm pretty happy with it ^^
Would it better to briefly describe the "dry" argument in the `scripts/styleR/use_styler.md`?

BTW, the default indent is 2 spaces for both (see https://styler.r-lib.org/reference/tidyverse_style.html and https://lintr.r-lib.org/reference/indentation_linter.html). And why not let the number of indent as a choice of 2 or 4 for users?

drosofff:
> I 've only used `lintr` on my code and I'm pretty happy with it ^^

Well, improvement often come from unhappy people, doesn't it ?

> Would it better to briefly describe the "dry" argument in the `scripts/styleR/use_styler.md`?

Yes, I did it. `--dry` is a pretty standard option, though.

> BTW, the default indent is 2 spaces for both (see https://styler.r-lib.org/reference/tidyverse_style.html and https://lintr.r-lib.org/reference/indentation_linter.html).

Yes ! actually it was specified in the [code](https://github.com/ARTbio/tools-artbio/blob/2b91edd5b401577b62906a013e43f9173eff9cac/scripts/styleR/styler.R#L21]


> And why not let the number of indent as a choice of 2 or 4 for users?

I agree that it's not a high priority, but it's easier to review code when you're familiar with its indentation.
Personally, I find the 4-indent more readable (standard in python world). I was not confortable with the indent 2 default for R in the IUC but I followed the guide line. Now the IUC switched to a 4-indent recommendation for R too. I am ok with that, and by the way I will change my Rstudio settings...

Thanks for the review @Ning-L !

Ning-L (approved):

PR #669, opened 2024-02-11
Manta minfix

drosofff:
Put profile as a macro
Always returns configManta.py.ini (suppress config_file_check variable) as a default dataset.
Update tests accordingly
Add Python 2 for the invocation of customConfigManta.py (otherwise the argparse module is not found)

Ed. The conda manta package is tortuous and complicated to use!


drosofff:
poke @bgruening 

A lot of commits and action for not much...  There is a non-boolean forced output now (the configfile).
Tell me if it fits your (minimal) expectations!

PR #698, opened 2024-11-05
Add Single Cell category

nomadscientist:
This tool I think can also work with pseudobulks / Single Cell data, so would be nice to have in the Galaxy Single Cell category.

### DIFF
--- a/tools/deseq2_normalization/.shed.yml
+++ b/tools/deseq2_normalization/.shed.yml
@@ -9,6 +9,7 @@ categories:
   - Transcriptomics
   - Sequence Analysis
   - Statistics
+  - Single Cell
 homepage_url: http://artbio.fr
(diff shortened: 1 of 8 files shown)


drosofff:
Hi @nomadscientist ,

Thank you for your PRs ! I understand your request and  I approve it.

However, as you may have noticed in the actions, there are linting as well as test (small) issues that are poping up with your changes, and you are not responsible for it.

Thus, for the sake of simplicity, I think I will open new PRs, one by tool, in which I will incorporate your changes for categories and bio.tools IDs, as well as the fixes for the linting issues.

Is that OK with you @nomadscientist ?

Thanks again for your attention !

PS. Same for the PR #699

nomadscientist:
This is awesome! Thank you so much!
And yes, I fear I've made that mistake a few times of putting multiple tool changes in the same PR

drosofff:
Closing the PR. Changes was done in separate tool-specific PRs.
Thanks @nomadscientist

